PR #1675, opened 2020-08-06
Improve ready filters spec

miemar:
Improved spec based on the suggestions on a daily call.


spec/features/awards_filters/ready_filter_spec.rb
```diff
@@ -6,25 +6,25 @@
   let!(:project_award_type) { (create :award_type, project: project) }
   let!(:verification) { create(:verification, account: owner) }
```
Alexey1100, line 8: 👍

I recommend the following structure for the spec:

```ruby
describe 'Ready Filter' do
  [1, 5, 10].each do |number_of_transfers|
    context "With #{number_of_transfers} transfers" do
      it "Doesn't duplicate transfers" do
        # ...
      end
    end
  end
end
```

PR #1678, opened 2020-08-13
Constellation Integration

Alexey1100:
This PR will include all necessary updates to support $DAG token transactions and Constellation Network.


app/lib/comakery/constellation.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def tx(hash)
+    JSON.parse(URI.open("https://#{explorer}/transactions/#{hash}").read)
+  end
+
+  private
+
+    def explorer_var_name
+      @explorer_var_name ||= "#{network.upcase}_BLOCK_EXPLORER_URL"
```
aquabu, line 22: Maybe this should start with `BLOCK_EXPLORER_URL_` so when sorting alphabetically it will group the env vars together.

app/models/concerns/constellation_addressable.rb
```diff
@@ -3,15 +3,28 @@ module ConstellationAddressable
...
+      if value !~ /^DAG\d[1-9A-HJ-NP-Za-km-z]{36}$/
         message = options[:message] || "should start with 'DAG', " \
           'followed by 37 characters'
         record.errors.add attribute, message
       end
     end
+
+    def validate_checksum(record, attribute, value)
+      included_checksum = value[3]&.to_i
+      computed_checksum = value[4..-1]&.scan(/\d/)&.map(&:to_i)&.reduce(&:+)&.modulo(9)
```
aquabu, line 22: is it possible that there would be no digits? if so we would end up with nil.modulo(9).

spec/models/account_spec.rb
```diff
@@ -174,6 +174,19 @@
...
+
+    it 'requires #constellation_wallet to be a valid constellation address' do
+      expect(account.constellation_wallet).to be_blank
+      expect(account).to be_valid
+      error_message = "Constellation wallet should start with 'DAG', followed by 37 characters"
+      error_message_checksum = 'Constellation wallet should include valid checksum'
+
+      expect(account.tap { |a| a.update(constellation_wallet: 'foo') }.errors.full_messages).to eq([error_message])
+      expect(account.tap { |a| a.update(constellation_wallet: '0x') }.errors.full_messages).to eq([error_message])
+      expect(account.tap { |a| a.update(constellation_wallet: "DAG#{'a' * 36}") }.errors.full_messages).to eq([error_message, error_message_checksum])
```
aquabu, line 186: May want to add a check for when there are no numbers in the address.

Alexey1100:
API requests and responses are still using following fields even for Constellation transactions:
- `ethereum_transaction_success`
- `ethereum_transaction_error`
- `ethereum_transaction_address`

It requires a bit of refactoring and renaming across the app, which is out of scope of this PR, I think, and should be updated along with these columns on Award model:
- `ethereum_contract_address`
- `ethereum_network`

.env.test
```diff
@@ -12,4 +12,5 @@ ETHEREUM_EXPLORER_SITE=fake-subdomain-for-tests.example.com
 SLACK_API_KEY=
 SLACK_API_SECRET=
 DISCORD_CLIENT_ID=
-DISCORD_CLIENT_SECRET=
\ No newline at end of file
+DISCORD_CLIENT_SECRET=
+BLOCK_EXPLORER_URL_CONSTELLATION_TESTNET=dummy
```
aquabu, line 16: Maybe block explorer URLs should be defined in application.rb and the environment variable overrides it if defined.

app/models/blockchain_transaction_award.rb
```diff
@@ -4,22 +4,34 @@ def update_transactable_status
...
+        case blockchain_transactable.transfer_type.name
+        when 'mint'
+          Comakery::Eth::Tx::Erc20::Mint.new(network, tx_hash)
+        when 'burn'
+          Comakery::Eth::Tx::Erc20::Burn.new(network, tx_hash)
+        else
+          Comakery::Eth::Tx::Erc20::Transfer.new(network, tx_hash)
+        end
+      else
+        Comakery::Eth::Tx.new(network, tx_hash)
```
aquabu, line 27: We might want to raise if we don't match to avoid bugs as we are adding more tokens.

aquabu: We can come back later to refactor the case statement templated tokens classes.

app/models/concerns/constellation_addressable.rb
```diff
@@ -0,0 +1,30 @@
...
+        validate_format(record, attribute, value)
+        validate_checksum(record, attribute, value)
+      end
+    end
+
+    def validate_format(record, attribute, value)
+      if value !~ /^DAG\d[1-9A-HJ-NP-Za-km-z]{36}$/
+        message = options[:message] || "should start with 'DAG', " \
+          'followed by 37 characters'
+        record.errors.add attribute, message
```
aquabu, line 16: During refactor should consider refactoring out the matcher and checksum function into the template.

PR #1691, opened 2020-09-22
Wallets & Balances

Alexey1100:


app/models/token.rb
```diff
@@ -101,6 +89,10 @@ def to_base_unit(amount)
     BigDecimal(10.pow(decimal_places || 0) * amount)&.to_s&.to_i
   end
 
+  def from_base_unit(amount)
+    BigDecimal(amount / 10.pow(decimal_places || 0))
```
aquabu, line 93: Might need to convert the numerator and denominator to BigDecimal before doing the division.

aquabu: Probably should do as a multiplication instead of a division.

app/models/wallet.rb
```diff
@@ -0,0 +1,13 @@
...
+  include BelongsToBlockchain
+
+  belongs_to :account
+  has_many :balances, dependent: :destroy
+
+  validates :address, :state, :source, presence: true
+  validates :address, blockchain_address: true
+  validates :_blockchain, uniqueness: { scope: :account_id }
+
+  enum state: { ok: 0, pending: 1, unclaimed: 2 }
```
aquabu, line 11: ok: 0, unclaimed: 1, pending:2

PR #1692, opened 2020-09-23
Sign-in redirect: Change logic

Riddlerrr:
To fix the bug https://app.asana.com/0/1138839033616420/1192251222542097

It happens because after sign in we redirects to `/tasks` path. But a new user doesn't have any tasks. So we redirect him back to the sign-in page.

### New redirection logic

- If you have >= 1 project on my projects page it redirects to /projects/mine
- If you have 0 projects and missions are not displayed (whitelabel) it redirects to /projects
- If you have 0 projects and missions are displayed (CoMakery.com) it redirects to / (this will show the same thing as the logged out home page on CoMakery.com)


Riddlerrr:
Oh, looks like there are still many tests to fix. Will do it

Riddlerrr:
Hmm, looks like it's only local error. CI is successfully done

app/controllers/sessions_controller.rb
```diff
@@ -74,7 +74,7 @@ def redirect_path
         @path
       else
         process_new_award_notice if current_account.new_award_notice
-        my_tasks_path
+        projects_path
```
aquabu, line 77: Just discussed this with @Alexey1100. Lets make this conditional:
* If you have >= 1 project on my projects page it redirects to `/projects/mine`
* If you have 0 projects and missions are not displayed (whitelabel) it redirects to `/projects`
* If you have 0 projects and missions are displayed (CoMakery.com) it redirects to `/` (this will show the same thing as the logged out home page on CoMakery.com)

This behavior works for whitelabel versions without missions on the home page.

Riddlerrr:
@aquabu @Alexey1100 Could you take a look of the new version?

app/controllers/sessions_controller.rb
aquabu: @Riddlerrr Is there a good way to get the cyclomatic complexity down?

aquabu (approved):

aquabu:
@Riddlerrr looks good. Might be good to get the cyclomatic complexity down - but not a blocking request.

aquabu:
@Alexey1100 can you review as well?

app/controllers/sessions_controller.rb
Riddlerrr: @aquabu You are right, better to refactor it =)
Done, looks much better now

app/controllers/sessions_controller.rb
```diff
@@ -64,22 +64,24 @@ def check_discord_oauth # rubocop:todo Metrics/CyclomaticComplexity
...
-      elsif current_account.projects.any?
-        process_new_award_notice if current_account.new_award_notice
-        my_project_path
-      elsif @whitelabel_mission
-        projects_path
-      else
-        root_path
-      end
+    def redirect_path
+      return redirect_to_my_tasks_with_notice if session[:redeem]
```
aquabu, line 68: Nice refactor :)

app/controllers/sessions_controller.rb
aquabu: Nice refactor in the latest commit :)

app/controllers/sessions_controller.rb
```diff
@@ -65,17 +65,23 @@ def check_discord_oauth # rubocop:todo Metrics/CyclomaticComplexity
...
-        my_tasks_path
-      elsif @path
-        @path
-      else
-        process_new_award_notice if current_account.new_award_notice
-        my_tasks_path
-      end
+      return redirect_to_my_tasks_with_notice if session[:redeem]
+      return @path if @path
+      return redirect_to_my_projects if current_account.projects.any?
```
Alexey1100, line 70: My Projects page also includes interested, team and archived projects, so I think we would want to redirect there if any of these are present.

(See `app/controllers/projects_controller.rb:20` for reference)

Alexey1100 (changes requested):

app/controllers/sessions_controller.rb
Riddlerrr: @Alexey1100 Sorry, maybe I didn't understand something.
I see interested and team projects on the `/projects/mine` page.

Do you mean not all of the projects contain in `current_account.projects`?

image

aquabu:
ok, this is getting more complicated. Let's go back to the previous functionality:
* When you login to CoMakery.com (not whitelabel) you end up on /tasks
* When you login to the whitelabel you are redirected to /projects/mine

Riddlerrr:
@Alexey1100 @aquabu Applied the last agreed logic, please take a look.

aquabu:
Looks good. Just waiting on @Alexey1100 review then will merge.

Alexey1100 (approved):

PR #1695, opened 2020-09-25
TransferController: Re-implement ordering in bypass ransack

Riddlerrr:
To solve the bug https://app.asana.com/0/1138839033616420/1184924517589713/f

Ransack has [an issue](https://github.com/activerecord-hackery/ransack#in-your-controller) with `distinct`.
You can use either distinct or ordering with joins, not at the same time.

After playing a bit with it I decided that the easiest way to solve it is to sort by the custom method and not rely on to ransack.

I've introduced `RansackReorder` concern which can order by model's columns and also support a special orders. We can add it like this:
https://github.com/CoMakery/comakery-server/blob/ad2766372874809a075da26816964956e628ad42/app/models/award.rb#L22

https://github.com/CoMakery/comakery-server/blob/ad2766372874809a075da26816964956e628ad42/app/models/award.rb#L118-L125


Riddlerrr:
@aquabu @Alexey1100 Guys, what do you think?

Riddlerrr:
There is a security issue. I believe we can fix it by introducing sorting whitelist params

app/controllers/dashboard/transfers_controller.rb
```diff
@@ -81,4 +81,17 @@ def transfer_params
         :transfer_type_id
       )
     end
+
+    def order_transfers(transfers)
```
Alexey1100, line 85: Looks good.
Would be perfect to extract that into a model concern, accepting custom params, so we can reuse it, and have some specs around it.

app/controllers/dashboard/transfers_controller.rb
```diff
@@ -81,4 +81,17 @@ def transfer_params
...
+      ordered_transfers = transfers.order('awards.created_at')
+      s = params.dig(:q, :s)
```
Alexey1100, line 87: Regarding the SQL injection warning, yeah I think we can avoid passing custom strings to query by introducing whitelist or by adding a case statement, which converts provided param to query.

Alexey1100 (changes requested):

app/controllers/dashboard/transfers_controller.rb
```diff
@@ -81,4 +81,17 @@ def transfer_params
...
+      return ordered_transfers unless s
```
aquabu, line 88: it would be more self documenting if the `s` param is assigned to a variable with a more descriptive name

app/models/award.rb
```diff
@@ -80,7 +82,7 @@ class Award < ApplicationRecord
   scope :in_progress, -> { where 'awards.status in(0,1,2,3)' }
   scope :contributed, -> { where 'awards.status in(1,2,3,4,5)' }
 
-  scope :transfer_ready, ->(blockchain) { accepted.joins(account: :wallets).where(wallets: { _blockchain: blockchain }).having('count(wallets) > 0') }
```
Riddlerrr, line 83: @Alexey1100  Is that having really needed? We already filtered cases without wallets by `inner join`.
I've removed it because it brings me errors.

Alexey1100: Let me check.
We have an additional requirement that account not only should have a wallet, but a wallet with specific blockchain.

Alexey1100: But if specs are passing then the join does job.

Alexey1100 (approved):

aquabu (approved):

PR #1696, opened 2020-09-25
Wallets & Balances

Alexey1100:


app/models/balance.rb
```diff
@@ -0,0 +1,11 @@
+class Balance < ApplicationRecord
+  belongs_to :wallet
+  belongs_to :token
+
+  validates :base_unit_value, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :wallet_id, uniqueness: { scope: :token_id }
```
aquabu, line 6: may want to verify that the token and the wallet belong to the same blockchain?

aquabu: if token has any balances then shouldn't be able to change the token's type or blockchain.

app/policies/wallet_policy.rb
```diff
@@ -0,0 +1,39 @@
...
+      if account
+        account.wallets
+      else
+        scope.none
+      end
+    end
+  end
+
+  def new?
+    true
```
aquabu, line 27: if account is not present then it should return false

aquabu (approved):

PR #1699, opened 2020-10-03
Adjust SimpleCov config to detect unused code better

Alexey1100:


spec/rails_helper.rb
```diff
@@ -4,6 +4,8 @@
 require 'simplecov'
 SimpleCov.start :rails do
   SimpleCov.minimum_coverage 98
+  SimpleCov.minimum_coverage_by_file 50
```
aquabu, line 7: nice - it might actually be easier if we set SlimpleCov.minimum_coverage = 100 as well. That way we wouldn't get unexpected failures when the total number of lines changes and throws off the ratio.

aquabu (approved):

PR #1703, opened 2020-10-13
Docs: Prettify response body json

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/175177557

I've configured `RspecApiDocumentation` to use `JSON.pretty_generate` as a `request_body_formatter`:
https://github.com/CoMakery/comakery-server/blob/b55ac11ca8cde6382d6ee0b0305dbc831b5bc607/config/initializers/rspec_api.rb#L80

Added docs generator script `bin/docs` for the convenient way to re-generate docs:
https://github.com/CoMakery/comakery-server/blob/b55ac11ca8cde6382d6ee0b0305dbc831b5bc607/bin/docs#L3

Also, all docs were updated.

Before:
II  Accounts API 2020-10-13 12-36-58

After:
II  Accounts API 2020-10-13 12-37-18

### DIFF
--- a/bin/docs
+++ b/bin/docs
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+
+RAILS_ENV=test bundle exec rake docs:generate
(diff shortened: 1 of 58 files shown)


Alexey1100 (approved):
Maybe we should add a note that the included signature is not valid, because formatting has been applied to the data after the signature was calculated.

PR #1708, opened 2020-10-14
Use the same Date of birth format across all API requests and responses

Alexey1100:


app/models/account.rb
```diff
@@ -72,7 +72,7 @@ class Account < ApplicationRecord
   # rubocop:enable Rails/UniqueValidationWithoutIndex
   validates :password, length: { minimum: 8 }, if: :password_required
   validates :first_name, :last_name, :country, :specialty, presence: true, if: :name_required
-  validates :date_of_birth, presence: { message: 'should be present in correct format (MM/DD/YYYY)' }, if: :name_required
+  validates :date_of_birth, presence: { message: 'should be present in correct format' }, if: :name_required
```
aquabu, line 75: I think we should inform them what the correct format is.

Alexey1100: Yes, we inform about the format in the API docs.

On the model level though rails automatically typecasts any standardised format into a date, as long as it parsable by `Date.parse` so the message wasn't correct.

aquabu (approved):

PR #1712, opened 2020-10-16
Wallet: Fix address validation with ore_id

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/175231932

I've found a bug in tests. Where were false positive because `address` implicitly auto-filled.

- tests fixed
- `address` can be null in the DB
- `state` now set `before_validation` and `on: :create`

P.S. CI fails because of https://github.com/CoMakery/comakery-server/pull/1711
After merge it should pass

### DIFF
--- a/app/models/concerns/ore_id_features.rb
+++ b/app/models/concerns/ore_id_features.rb
@@ -2,7 +2,7 @@ module OreIdFeatures
   extend ActiveSupport::Concern
 
   included do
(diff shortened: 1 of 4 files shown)


Alexey1100:
Nice. There should be some conflicts with #1714 though, since wallet specs significantly refactored there, let me try to merge them.

aquabu (approved):

PR #1714, opened 2020-10-16
OreId Integration

Alexey1100:


db/schema.rb
```diff
@@ -332,6 +332,14 @@
     t.index ["token_id"], name: "index_missions_on_token_id"
   end
 
+  create_table "ore_ids", force: :cascade do |t|
```
aquabu, line 335: to avoid rails model naming confusion, maybe we should call this `aikon_accounts` and the foreign key would be `aikon_account_id`

Alexey1100: I propose `OreAccount` model, with the table `ore_accounts` and key `ore_account_id`

aquabu: that works 👍

app/models/account.rb
```diff
@@ -47,6 +47,7 @@ class Account < ApplicationRecord
   has_many :account_token_records_synced, -> { where synced: true } # rubocop:todo Rails/InverseOf
   has_many :wallets, dependent: :destroy
   has_many :balances, through: :wallets
+  has_one :ore_id, dependent: :destroy
```
aquabu, line 50: ore_account instead of ore_id so we get ore_account_id instead of ore_id_id

app/models/ore_id_service.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  private
+
+    def create_remote_params
+      {
+        'name' => account.name,
+        'user_name' => account.nickname || '',
+        'email' => account.email,
+        'picture' => account.image ? Refile.attachment_url(account, :image) : '',
+        'user_password' => SecureRandom.hex(32) + '!',
```
aquabu, line 63: should make a comment here that this password should not be stored in the DB for future maintenance.

aquabu (approved):

PR #1716, opened 2020-10-19
Add refresh transfer rules button

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/175048822

### How it  works

The main code placed in `TransferRulesSyncJob` which run from the controller on refresh.

In the job we make an Etherium API call to filter all [setAllowGroupTransfer](https://github.com/CoMakery/comakery-security-token/blob/master/contracts/RestrictedToken.sol#L262) function call for contract address.

After that, we process every event and initialize new `TransferRules` and create new RegGroup if needed.
In the processing we have to decode reg groups (from `topics`, because it indexed) and `lockup_until` (from `data`).

Note: we can't update `lockup_until` for existing TransferRules, so we have to create new. And the old one will be deleted on creating.

In the end we save all initialized `TransferRules`.

### How does it look
The new button places on the "transfer rules" page. Which available for comakery token only.
WL 2020-10-23 12-43-53

### After merge

Don't forget to add `rails blockchain_sync:transfer_rules` to Heroku Scheduler after merge


Alexey1100:
The task might get quite tricky, because there's no simple way to detect new transfer rules created outside Comakery.
We discussed some ways to do it, previously:
 - A background job to listen for contract events
 - Getting all possible combination of created Reg Groups and polling them to check whether a Transfer Rule is present 

I'm not sure, if this is required for the task though.

app/jobs/blockchain_job/comakery_security_token_job/transfer_rules_sync_job.rb
```diff
@@ -0,0 +1,84 @@
...
+        raise 'Token is not Comakery Type' unless @token._token_type_comakery_security_token?
+
+        @decoder = Ethereum::Decoder.new
+        @transfer_rules = []
+        @reg_groups = @token.reg_groups.to_a
+
+        @client = Comakery::Eth.new(@token.blockchain.explorer_api_host).client
+
+        filtered_events.each { |event| process_event(event) }
+        @transfer_rules.each(&:save!)
```
aquabu, line 17: Maybe wrap this in a transaction so that updates are atomic and catch any failures on `save!` to rollback.

aquabu (approved):

PR #1719, opened 2020-10-26
RegGroup: Make blockchain_id readonly

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/175413599

RegGroup#blockchain_id became read-only and auto-generated for new RegGroups

WL 2020-10-26 14-02-35

It works both front-end and back-end parts

Also, I've changed API docs because `blockchain_id` is not required anymore but you still possible to set your own number.

### DIFF
--- a/spec/acceptance/api/v1/8_reg_groups_spec.rb
+++ b/spec/acceptance/api/v1/8_reg_groups_spec.rb
@@ -73,17 +73,14 @@ resource 'VIII. Reg Groups' do
 
     with_options scope: :reg_group, with_example: true do
       parameter :name, 'reg group name', required: true, type: :string
-      parameter :blockchain_id, 'reg group id on blockchain', required: true, type: :string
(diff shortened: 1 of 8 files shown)


spec/acceptance/api/v1/8_reg_groups_spec.rb
```diff
@@ -73,17 +73,14 @@
 
     with_options scope: :reg_group, with_example: true do
       parameter :name, 'reg group name', required: true, type: :string
-      parameter :blockchain_id, 'reg group id on blockchain', required: true, type: :string
+      parameter :blockchain_id, 'reg group id on blockchain', type: :string
```
aquabu, line 76: should mention in the API that the blockchain_id will be autogenerated on create and on update you won't be able to modify blockchain_id

aquabu (approved):

PR #1723, opened 2020-10-28
TransfersController API: Set description's default to empty string

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/175360834

Just changed default:
https://github.com/CoMakery/comakery-server/blob/81a8c17073a9a8dc9b91b83fc290f039f9bcd3ee/app/controllers/api/v1/transfers_controller.rb#L31

### DIFF
--- a/spec/controllers/api/v1/transfers_controller_spec.rb
+++ b/spec/controllers/api/v1/transfers_controller_spec.rb
@@ -84,6 +84,19 @@ RSpec.describe Api::V1::TransfersController, type: :controller do
         post :create, params: params
         expect(response).to have_http_status(:created)
       end
+
+      it 'fills default values for why, requirements and description' do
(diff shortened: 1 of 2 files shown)


spec/controllers/api/v1/transfers_controller_spec.rb
```diff
@@ -84,6 +84,19 @@
...
+
+      it 'fills default values for why, requirements and description' do
+        params = build(:api_signed_request, { transfer: valid_attributes.except(:why, :requirements, :description) }, api_v1_project_transfers_path(project_id: project.id), 'POST')
+        params[:project_id] = project.id
+
+        post :create, params: params
+        expect(response).to have_http_status(:created)
+
+        award = Award.last
+        expect(award.why).to eq '—'
```
Alexey1100, line 96: Maybe we should set these two to empty as well.

Riddlerrr: ok, no problem =)

Alexey1100 (approved):

aquabu (approved):

PR #1725, opened 2020-10-28
Create Synchronisation model and concern

Alexey1100:
Define reusable Syncrhonisable Concern to control syncrhonisation state


db/migrate/20201028102447_create_synchronisations.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSynchronisations < ActiveRecord::Migration[6.0]
+  def change
+    create_table :synchronisations do |t|
+      t.references :synchronisable, polymorphic: true, index: { name: 'idx_syncs_on_sync_type_and_sync_id' }
+      t.integer :status
```
Alexey1100, line 5: default 0

app/models/concerns/synchronisable.rb
```diff
@@ -0,0 +1,52 @@
+module Synchronisable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :synchronisations, as: :synchronisable, dependent: :destroy
+
+    def min_seconds_between_syncs
+      0
```
Alexey1100, line 8: 0.01

db/schema.rb
```diff
@@ -452,6 +452,15 @@
     t.string "name"
   end
 
+  create_table "synchronisations", force: :cascade do |t|
+    t.string "synchronisable_type"
+    t.bigint "synchronisable_id"
+    t.integer "status"
```
Riddlerrr, line 458: Should it be `null: false`?

PR #1727, opened 2020-10-30
Add ability to hide testnets blockchains

Riddlerrr:
Tasks: https://www.pivotaltracker.com/story/show/175469614 and https://www.pivotaltracker.com/story/show/175483264

I've decided to always show the full list of blockchains for `comakery_admin`'s. Not sure if it correct.:
https://github.com/CoMakery/comakery-server/blob/4304400255de7eda9a2fc40cd463596e8e6a6f68/app/models/wallet.rb#L32

Spotify Premium 2020-10-30 15-27-49

Spotify Premium 2020-10-30 15-13-28


Alexey1100:
🤔
I think the list of the blockchains for Token and Wallet should be definitely the same, regardless account status, because a wallet using a testnet doesn't have a purpose without a token with the same network.

Which means we can extract all the blockchain availability logic into the Blockchain class.

app/models/concerns/belongs_to_blockchain.rb
```diff
@@ -17,6 +17,14 @@ def self.blockchain_for(name)
       "Blockchain::#{name.camelize}".constantize.new
     end
 
+    def self.testnets_available?
```
Alexey1100, line 20: Looks like duplicate of

https://github.com/CoMakery/comakery-server/blob/d79dfc56ba84f82f971f5743c313e51f010d0f0d/app/models/blockchain.rb#L27

Riddlerrr: Right 👍 , forgot to delete it

app/models/wallet.rb
```diff
@@ -16,12 +16,17 @@ class Wallet < ApplicationRecord
...
+    available_blockchains.map(&:key) - account.wallets.pluck(:_blockchain)
   end
 
   private
 
     def ore_id_and_pending?
       pending? && ore_id?
     end
+
+    def testnet_blockchains_available?
```
aquabu, line 29: remove?

spec/models/blockchain_spec.rb
```diff
@@ -7,6 +7,24 @@
...
+      is_expected.to include Blockchain::BitcoinTest
+      is_expected.to include Blockchain::Bitcoin
+    end
+
+    it 'do not returns testnets if TESTNETS_AVAILABLE set to false' do
+      ENV['TESTNETS_AVAILABLE'] = 'false'
+
+      is_expected.not_to include Blockchain::BitcoinTest
+      is_expected.to include Blockchain::Bitcoin
+    end
```
aquabu, line 26: may need to reset the ENV variable to what it started as after each test

probably should write a test to confirm null testnet value has the same result as true value

aquabu (approved):

PR #1729, opened 2020-11-02
Update Transfers API docs. Don't allow to delete ore_id sourced wallets.

Alexey1100:
API docs updates (https://www.pivotaltracker.com/story/show/175562603)
- Remove deprecated 'source' param
- Describe default 'transfer_type_id'
- Regenerate docs

Wallet updates (https://www.pivotaltracker.com/story/show/175360769, https://www.pivotaltracker.com/story/show/175522710)
- Throw an abort on the model level when destroying an OreId sourced wallet
- Handle abort in dashboard and api controllers
- Show wallet state instead of addres if address is not present
- Don't show remove button if wallet can not be removed


Alexey1100:
Closes:
https://www.pivotaltracker.com/story/show/175360769
https://www.pivotaltracker.com/story/show/175562603

aquabu (approved):

PR #1730, opened 2020-11-02
Introduce Algorand standard Asset token type

Riddlerrr:
Task https://www.pivotaltracker.com/story/show/175518070

I've changed `contract address` to `asset id` in the form:
CoMakery - Achieve Big Missions 2020-11-03 13-49-52

But I still store it in `Token.contract_address` in the DB.


app/controllers/tokens_controller.rb
```diff
@@ -76,6 +76,10 @@ def fetch_contract_details
     when /^[a-fA-F0-9]{40}$/
       qtum = Comakery::Qtum.new(host)
       @symbol, @decimals = qtum.fetch_symbol_and_decimals(params[:address])
+    when /^[a-zA-Z0-9]{58}$/ # AlgorandTest
```
Alexey1100, line 79: Since it's not a smart contract we don't need to dynamically fetch/update symbol and decimals.

app/javascript/components/TokenForm.jsx
```diff
@@ -316,7 +316,7 @@ class TokenForm extends React.Component {
               symbolLimit={0}
             />
 
-            {this.state['token[_token_type]'].match(/erc20|qrc20|comakery_security_token/) &&
+            {this.state['token[_token_type]'].match(/erc20|qrc20|comakery_security_token|asa/) &&
```
Alexey1100, line 319: Not a smart contract

app/models/token_type/asa.rb
```diff
@@ -0,0 +1,88 @@
+class TokenType::Asa < TokenType
+  # Generated template for implementing a new token type subclass
+  # See: rails g token_type -h
+
+  # Name of the token type for UI purposes
+  # @return [String] name
+  def name
+    'ASA'
```
Alexey1100, line 8: I believe name should be Algo, symbol ALGO

Alexey1100: @aquabu should we first implement ALGO coin or we doing Algorand smart contracts?

app/models/token_type/asa.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  # Symbol of the token type for UI purposes
+  # @return [String] symbol
+  def symbol
+    'ASA'
```
Alexey1100, line 14: I think we can integrate dynamic symbol and decimals inside the subclass itself.
Same as with erc20/qrc20:

https://github.com/CoMakery/comakery-server/blob/5eb61e29fdae6d8d2b673d5ef4e298905982b09b/app/models/token_type/erc20.rb#L14

app/controllers/tokens_controller.rb
```diff
@@ -67,16 +67,20 @@ def update
...
 
-    case params[:address]
-    when /^0x[a-fA-F0-9]{40}$/
-      web3 = Comakery::Web3.new(host)
-      @symbol, @decimals = web3.fetch_symbol_and_decimals(params[:address])
-    when /^[a-fA-F0-9]{40}$/
-      qtum = Comakery::Qtum.new(host)
-      @symbol, @decimals = qtum.fetch_symbol_and_decimals(params[:address])
-    end
+    client =
```
Alexey1100, line 74: 👍

Later we can refactor all clients into a single interface (probably inside the token type subclass) and get rid of the case condition.

app/lib/comakery/algorand.rb
```diff
@@ -0,0 +1,24 @@
...
+  def symbol
+    asset_details.dig('asset', 'params', 'unit-name')
+  end
+
+  def decimals
+    asset_details.dig('asset', 'params', 'decimals')
+  end
+
+  def asset_details
+    @asset_details ||= get_request('/v2/assets/', @asset_id)
```
Alexey1100, line 17: Wondering if we can move `/v2/assets` into `Blockchain::Algorand#url_for_asset_api(id)`.

So all the urls will be contained inside the subclass.
And the wrapper will just call `@blockchain.url_for_asset_api(@asset_id)`

app/models/token_type/asa.rb
```diff
@@ -0,0 +1,103 @@
...
+  # Number of decimals
+  # @return [Integer] number
+  def decimals
+    @decimals ||= contract&.decimals&.to_i
+  end
+
+  # Wallet logo filename for UI purposes (relative to `app/assets/images`)
+  # @return [String] filename
+  def wallet_logo
+    nil
```
Alexey1100, line 26: We can add OreId logo here :)

https://aikon.com/static/media/ORE-ID.18a7f78c.zip

aquabu (approved):

Riddlerrr:
@aquabu It is ready to merge now

PR #1731, opened 2020-11-03
Improvements for TESTNETS_AVAILABLE env

Riddlerrr:
Some improvement with TESTNETS_AVAILABLE env after https://github.com/CoMakery/comakery-server/pull/1727

### DIFF
--- a/spec/models/blockchain_spec.rb
+++ b/spec/models/blockchain_spec.rb
@@ -10,25 +10,24 @@ describe Blockchain, type: :model do
 
   describe '#available' do
     subject { described_class.available }
-    after { ENV['TESTNETS_AVAILABLE'] = 'true' }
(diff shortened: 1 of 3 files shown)


spec/models/blockchain_spec.rb
```diff
@@ -10,6 +10,7 @@
 
   describe '#available' do
     subject { described_class.available }
+    after { ENV['TESTNETS_AVAILABLE'] = 'true' }
```
Alexey1100, line 13: Still doesn't caches the variable before the run.

Alternatively, we can use stubbing:
`allow(ENV).to receive(:[]).with("TESTNETS_AVAILABLE").and_return("true")`

aquabu (approved):

PR #1734, opened 2020-11-04
Implement Ore ID password reset url creation

Alexey1100:
https://www.pivotaltracker.com/story/show/175195123

- Create a new auth token
- Return an actual Ore ID reset password url with valid params and auth
token

### DIFF
--- a/app/models/ore_id_service.rb
+++ b/app/models/ore_id_service.rb
@@ -48,8 +48,27 @@ class OreIdService
     raise OreIdService::RemoteInvalidError
   end
 
+  def create_token
+    response = handle_response(
+      self.class.post(
+        '/app-token',
+        headers: request_headers
+      )
+    )
+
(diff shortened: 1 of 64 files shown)


app/models/ore_id_service.rb
```diff
@@ -48,8 +48,27 @@ def permissions
...
+  end
+
   def password_reset_url(redirect_url)
-    "https://example.org?redirect=#{redirect_url}"
+    params = {
+      app_access_token: create_token,
+      provider: :email,
+      callback_url: redirect_url,
+      background_color: 'FFFFFF',
+      state: ''
```
Alexey1100, line 68: `state` is an optional string which will be included in the callback params

we can add some metadata here (e.g. `managed_account_id`) if needed

aquabu (approved):

PR #1736, opened 2020-11-05
AccountsController: Show status and last synced data

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/175048825

Actually, there are no problems in the backend part.
Since we can't update `AccountTokenRecord` fields (readonly) we creating a new record with status 'created' and after that we trying to sync it. On the page we just show the last record.

From the user's side we definitely updated the rows and it looks like everything is ok and that's confusing.

I've decided just add the status column to the table and show there the actual info and that data was changed:
CoMakery - Achieve Big Missions 2020-11-06 13-24-09

### DIFF
--- a/app/views/dashboard/accounts/_account.html.erb
+++ b/app/views/dashboard/accounts/_account.html.erb
@@ -1,11 +1,13 @@
 <% account = account.decorate %>
-<% record = account.account_token_records.where(token: @project.token).last %>
-<% form_url = record ? project_dashboard_account_path(@project, record) : root_path %>
-<% form_data = record ? { controller: "account-form", target: "account-form.form", 'account-form-address' => account.address_for_blockchain(@project.token._blockchain), 'account-form-account-id' => account.id, 'account-form-account-token-records-path' => api_v1_project_account_token_records_path(project_id: @project.id), 'account-form-transactions-path' => api_v1_project_blockchain_transactions_path(project_id: @project.id)}.merge(@project.token.decorate.eth_data('account-form')) : nil %>
+<% synced_record = account.account_token_records.where(token: @project.token, status: :synced).last %>
+<% last_record = account.account_token_records.where(token: @project.token).last %>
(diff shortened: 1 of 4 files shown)


Alexey1100:
Yep, the problem on the UI side only, since it wasn't created with such multi-step synchronisation logic in mind.
I believe we wanted to improve it at some point.

Riddlerrr:
I tried to show all records but it become too complicated. So I decided to just show last synced data was updated in the same row and cross it.
CoMakery - Achieve Big Missions 2020-11-06 13-24-09

app/views/dashboard/accounts/_account.html.erb
```diff
@@ -1,16 +1,18 @@
...
+<% last_record_is_not_synced = synced_record && last_record && synced_record.id != last_record.id %>
+<% form_url = last_record ? project_dashboard_account_path(@project, last_record) : root_path %>
+<% form_data = last_record ? { controller: "account-form", target: "account-form.form", 'account-form-address' => account.address_for_blockchain(@project.token._blockchain), 'account-form-account-id' => account.id, 'account-form-account-token-records-path' => api_v1_project_account_token_records_path(project_id: @project.id), 'account-form-transactions-path' => api_v1_project_blockchain_transactions_path(project_id: @project.id)}.merge(@project.token.decorate.eth_data('account-form')) : nil %>
 
-<%= form_with model: record, url: form_url, data: form_data, class: 'account-form' do |f| %>
-	
-	<% if policy(@project).edit_accounts? && record && account.address_for_blockchain(@project.token._blockchain) %>
+<%= form_with model: last_record, url: form_url, data: form_data, class: 'account-form' do |f| %>
+
+	<% if policy(@project).edit_accounts? && last_record && account.address_for_blockchain(@project.token._blockchain) %>
```
aquabu, line 10: looks like mixed tabs and spaces. we can setup rubocop to check for this.

aquabu (approved):

PR #1739, opened 2020-11-10
Introduce ORE ID authorisation logic and ALGO transaction signature

Alexey1100:
Implements:

https://www.pivotaltracker.com/story/show/175580920
https://www.pivotaltracker.com/story/show/175561400


app/models/ore_id_service.rb
```diff
@@ -59,18 +59,50 @@ def create_token
...
+    issuer = transfer.issuer
+    recipient = transfer.account
+    issuer_ore_id = issuer.ore_id_account
+    blockchain = transfer.token.blockchain
+    issuer_wallet = issuer.wallets.find_by!(_blockchain: blockchain.key)
+    recipient_wallet = recipient.wallets.find_by!(_blockchain: blockchain.key)
+    transaction_data = {
+      from: issuer_wallet.address,
+      to: recipient_wallet.address,
+      amount: (transfer.amount * 1000000).to_i, # in microalgos
```
Alexey1100, line 86: When operating with blockchain transaction here, amount will be calculated automatically with proper decimals

app/controllers/auth/ore_id_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Auth::OreIdController < ApplicationController
+  include OreIdCallbacks
+  skip_after_action :verify_authorized
```
Alexey1100, line 3: TODO: Policy for new and receive

app/controllers/concerns/ore_id_callbacks.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    def verify_errorless
+      if received_error
+        flash[:error] = received_error
+        redirect_to received_state[:redirect_back_to]
+      end
+    end
+
+    def verify_received_account
+      head 401 unless current_account.id == received_state['account_id']
```
Alexey1100, line 54: Probably will be moved to policy

app/controllers/sign/ore_id_controller.rb
```diff
@@ -1,28 +1,32 @@
...
-    # Should accept transaction instead of transfer
-
     transfer = Award.find(params.require(:transfer_id))
-    head 401 if current_account != transfer.issuer
+    source_wallet = current_account.wallets.find_by!(_blockchain: transfer.token._blockchain).address
+
+    transaction = BlockchainTransactionAward.create!(
+      blockchain_transactable: transfer,
+      source: source_wallet,
+      nonce: 1
```
Alexey1100, line 13: I believe nonce is not needed here, since we are not generating raw transaction.

app/controllers/sign/ore_id_controller.rb
```diff
@@ -1,28 +1,32 @@
...
   include OreIdCallbacks
+  skip_after_action :verify_authorized
 
   # GET /sign/ore_id/new
   def new
-    # Should accept transaction instead of transfer
-
     transfer = Award.find(params.require(:transfer_id))
-    head 401 if current_account != transfer.issuer
+    source_wallet = current_account.wallets.find_by!(_blockchain: transfer.token._blockchain).address
```
Alexey1100, line 8:
```suggestion
    source_wallet = current_account.address_for_blockchain(transfer.token._blockchain)
```

app/controllers/sign/ore_id_controller.rb
```diff
@@ -1,28 +1,32 @@
...
-    raise OreIdCallbacks::NoStateError unless params[:state]
+    verify_errorless
+    verify_received_account
 
-    head 401 unless current_account.id == received_state['account_id']
+    transaction = BlockchainTransactionAward.find(received_state['transaction_id'])
+    transaction.tx_hash = params.require(:transaction_id)
+    transaction.tx_raw = Base64.decode64(params.require(:signed_transaction))
+    transaction.status = :pending
+    transaction.save!
```
Alexey1100, line 28: I think we can reuse the same pattern here as in BlockchainTransactionsController:

```ruby
    if transaction.update(tx_hash: params.require(:transaction_id), tx_raw: Base64.decode64(params.require(:signed_transaction)))
      transaction.update_status(:pending)
      BlockchainJob::BlockchainTransactionSyncJob.perform_later(transaction)
    end
```

And for the error state we would need something like this:

```ruby
transaction.update_status(:cancelled, received_error)
```

app/views/shared/_transfer_button.html.erb
```diff
@@ -11,34 +11,42 @@
 
     <% elsif transfer.recipient_address.blank? && current_account == transfer.account %>
       <%= link_to 'needs wallet', account_path  %>
-    
+    <% elsif transfer.token&.blockchain&.supported_by_ore_id? && (transfer.issuer&.ore_id_account.nil? || transfer.account.ore_id_account.nil?)%>
```
Riddlerrr, line 14: We have to change the condition. 
A receiver can don't have `ore_id_account`, just added wallet.

Alexey1100: Yep, and that should be only possible for Algo transfers, since Asa will require an Ore ID account even on receiver side.

Alexey1100: That might be confusing..

Riddlerrr: > since Asa will require an Ore ID account even on receiver side.

Why?! You can add ASA to any Algorand wallet, isn't it?

Alexey1100: Technically it's possible, but we will have to instruct user to create such approval transaction externally with `assetId` from a token on CoMakery and them submit tx hash or flip some switch on UI, so we can track the allowance on our side.

It feels like it eliminates all the simplicity we rely on OreId for.

app/models/ore_id_service.rb
```diff
@@ -93,6 +109,16 @@ def request_headers
       }
     end
 
+    def algo_transfer_transaction(transaction)
+      {
+        from: transaction.source,
+        to: transaction.destination,
+        amount: transaction.amount,
+        note: "CoMakery payment for Transaction##{transaction.id}",
```
aquabu, line 117: no note

aquabu (approved):

PR #1745, opened 2020-11-23
Sync Algorand's transactions

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/175845429

In this PR I've added sync for ALGO transactions only.
Assets transactions will be implemented separately.

### DIFF
--- a/app/lib/comakery/algorand/tx.rb
+++ b/app/lib/comakery/algorand/tx.rb
@@ -18,7 +18,47 @@ class Comakery::Algorand::Tx
     transaction_data&.fetch('confirmed-round', nil)
   end
 
-  def confirmed?(_number_of_confirmations = 1)
-    confirmed_round.present?
+  def current_round
+    data.fetch('current-round', 0)
   end
+
+  def closing_amount
+    transaction_data.fetch('closing-amount', nil)
+  end
+
+  def sender_address
+    transaction_data.fetch('sender', '')
+  end
+
+  def receiver_address
+    transaction_data.dig('payment-transaction', 'receiver') || ''
+  end
+
+  # In MicroAlgos for ALGO transaction
+  def amount
+    transaction_data.dig('payment-transaction', 'amount') || 0
+  end
+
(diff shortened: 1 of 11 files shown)


app/models/blockchain/algorand.rb
```diff
@@ -24,6 +24,10 @@ def asset_api_path(asset_id)
     "/v2/assets/#{asset_id}"
   end
 
+  def get_transaction_path(tx_hash)
```
Alexey1100, line 27: Looks like we already have it:

Riddlerrr: 👍

app/lib/comakery/algorand/tx.rb
```diff
@@ -0,0 +1,64 @@
...
+    transaction_data.dig('payment-transaction', 'receiver') || ''
+  end
+
+  # In MicroAlgos for ALGO transaction
+  def amount
+    transaction_data.dig('payment-transaction', 'amount') || 0
+  end
+
+  def confirmed?(_number_of_confirmations = 1)
+    confirmed_round.present? && current_round > confirmed_round
```
aquabu, line 43: perhaps `current_round >= confirmed_round`
number of confirmations is ignored due to blockchain finality on Algorand?

app/lib/comakery/algorand/tx.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def valid?(blockchain_transaction)
+    return false unless confirmed?
+    return false unless valid_addresses?(blockchain_transaction)
+    return false unless valid_amount?(blockchain_transaction)
+
```
aquabu, line 50: May want an additional check here to verify the database stored blockchain_transaction current block is < confirmed round on the blockchain transaction. This is to verify that the transaction is is in a later block on the blockchain then when the payment interaction started on CoMakery.

aquabu (approved):

PR #1746, opened 2020-11-24
Improve OreId auth logic, Prevent manual creation of Algorand wallets

Alexey1100:
Implements:

https://www.pivotaltracker.com/story/show/175863184
https://www.pivotaltracker.com/story/show/175863153

### DIFF
--- a/app/controllers/sign/ore_id_controller.rb
+++ b/app/controllers/sign/ore_id_controller.rb
@@ -1,7 +1,7 @@
 class Sign::OreIdController < ApplicationController
   include OreIdCallbacks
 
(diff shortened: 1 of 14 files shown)


app/controllers/sign/ore_id_controller.rb
```diff
@@ -15,8 +15,8 @@ def new
 
   # GET /sign/ore_id/receive
   def receive
-    verify_errorless
-    verify_received_account
+    verify_errorless or return
```
aquabu, line 18: double check you want `or` and not `||`

aquabu (approved):

PR #1748, opened 2020-11-26
Algorand Asset Opt-in

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/175822817


app/models/blockchain_transaction.rb
```diff
@@ -107,7 +107,6 @@ def populate_data # rubocop:todo Metrics/CyclomaticComplexity
       self.network ||= token._blockchain if token._token_type_on_ethereum? || token._token_type_dag? || token._token_type_algo? || token._token_type_asa?
 
       self.current_block ||= Comakery::Eth.new(token.blockchain.explorer_api_host).current_block if token._token_type_on_ethereum?
-      self.current_block ||= Comakery::Algorand.new(token.blockchain).last_round if token._token_type_algo? || token._token_type_asa?
```
Alexey1100, line 110: Looks like it's moved to `blockchain_transaction_award.rb`, while it's supposed to be here, I guess.

app/policies/token_opt_in_policy.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def initialize(account, token_opt_in)
+    @account = account
+    @token_opt_in = token_opt_in
+  end
+
+  def create?
+    @token_opt_in.new_record? &&
+      @token_opt_in.valid? &&
+      @account.wallets.where(id: @token_opt_in.wallet_id).exists?
```
Alexey1100, line 12: I think the first two conditions are not related to policing 🤔, since invalid record should raise it's own Error.

config/rubocop/default.yml
```diff
@@ -21,7 +21,7 @@ AllCops:
 # Low cyclomatic complexity is the simplest indication of code quality, readability and maintainability.
 # Bypassing this check should trigger a team discussion
 Metrics/CyclomaticComplexity:
-  Max: 5
+  Max: 6
```
Alexey1100, line 24: Do we want to alter that?

Alexey1100:
Seems clear  : )
I'm ready to integrate that into #1749

app/policies/token_opt_in_policy.rb
Alexey1100: And I think we need to add a simple policy spec here.

config/rubocop/default.yml
aquabu: Discussed

aquabu (approved):

PR #1752, opened 2020-12-02
Logging API requests

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/175402875

After successful API's signature verifying we will create a new ApiRequestLog record with these attributes:
https://github.com/CoMakery/comakery-server/blob/3612e89b4caa3774a5585a0a1ff4cd24a0c0e7a4/db/migrate/20201202070221_create_api_request_logs.rb#L4-L7


Alexey1100 (approved):
👍

aquabu (approved):

PR #1756, opened 2020-12-06
Fixes to make deploying a new server instance easier

aquabu:


Gemfile
```diff
@@ -127,7 +127,7 @@ gem 'barnes', '~> 0.0.7'
 
 gem 'bootsnap', '~> 1.4'
 
-gem 'sentry-raven', '~> 2.12'
+# gem 'sentry-raven', '~> 2.12' # disabled because sentry is down and it blocks deploying new instances
```
aquabu, line 130: uncomment this now that sentry is fixed on heroku

Gemfile
```diff
@@ -154,3 +154,5 @@ gem 'validate_url', '~> 1.0'
 gem 'overcommit', '~> 0.53.0'
 
 gem 'fuubar', '~> 2.5'
+
+gem 'bugsnag', '~> 6.18'
```
aquabu, line 158: remove now that we are switching back to sentry

Procfile
```diff
@@ -1,3 +1,2 @@
 web: bundle exec puma -C config/puma.rb
 worker: bundle exec sidekiq -v -C ./config/sidekiq.yml
-release: bin/rails db:migrate
```
aquabu, line 3: revert this

config/initializers/env_verifier.rb
```diff
@@ -2,5 +2,5 @@
   var = line.split('=')&.first
 
   next unless var.present?
-  raise "Missing required ENV variable: #{var}" if ENV[var].blank?
+  Rails.logger.info "Missing required ENV variable: #{var}" if ENV[var].blank?
```
aquabu, line 5: May need some more research on how to handle this. Need to:
* block deployment if you don't have the ENV var
* but env vars are only updated on Heroku after initial app setup
* result: this is fine if the app is already deployed but not on initial deployment?

aquabu:
This is outdated now. We merged the appropriate changes in another commit.

PR #1758, opened 2020-12-07
Add bugsnag gem to use with heroku Bugsnag add-on

Alexey1100:


aquabu:
I merged this into `bug/new-deploy` instead. So I cam going to close this PR.

PR #1766, opened 2020-12-10
Use `db:prepare` instead of `db:migrate`

Riddlerrr:
Tasks: 
- https://www.pivotaltracker.com/story/show/176006709
- https://www.pivotaltracker.com/story/show/176029688

Rails 6 introduced [db:prepare](https://blog.saeloun.com/2019/04/11/rails-6-rails-db-prepare.html) which we can use instead of `db:migrate`. It's safe to run even in case DB was not created.

`db:prepare` logic:
- if DB does not exists it will run `db:setup`. So it creates the DB and load schema and seeds
- if DB exists it just run `db:migrate`

So in most cases it is just like `db:migrate` call.

### DIFF
--- a/app.json
+++ b/app.json
@@ -6,7 +6,7 @@
   "logo": "https://avatars0.githubusercontent.com/u/17172662",
   "success_url": "/",
   "scripts": {
-    "postdeploy": "bundle exec rails db:migrate"
+    "postdeploy": "bundle exec rails db:prepare"
   },
   "buildpacks": [
     {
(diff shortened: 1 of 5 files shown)


Alexey1100 (approved):
👍

app.json
```diff
@@ -6,7 +6,7 @@
   "logo": "https://avatars0.githubusercontent.com/u/17172662",
   "success_url": "/",
   "scripts": {
-    "postdeploy": "bundle exec rails db:migrate"
+    "postdeploy": "bundle exec rails db:prepare"
```
aquabu, line 9: add `date:migrate` to postdeploy

aquabu (approved):

PR #1769, opened 2020-12-11
Algorand opt ins: Add availability to opt-in into algorand security token

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/176114032

Specificity:
- App Opt-in transaction can't send second time without error.


app/lib/comakery/algorand/tx/app.rb
```diff
@@ -0,0 +1,27 @@
...
+    transaction_data.dig('application-transaction', 'application-id')
+  end
+
+  def receiver_address
+    nil
+  end
+
+  # In the minimal App token unit
+  # amount 100 mean 1.00 when App decimal is 2
+  def amount
```
aquabu, line 20: should either delete or it should be nil depending on inheritance

app/models/ore_id_service.rb
```diff
@@ -168,9 +168,8 @@ def app_transaction(transaction)
         type: 'appl',
         appIndex: transaction.token.contract_address.to_i,
         to: nil,
-        amount: nil
-
-        # TODO: Add `appArgs:` to pass method name and params
+        amount: nil,
+        appOnComplete: 1 # OptIn type
```
aquabu, line 172: not in all call types

aquabu (approved):

PR #1771, opened 2020-12-16
Update synchronisable rescheduling logic

Alexey1100:
- Allow next sync at current_time instead of zero, for instances without
synchronisations yet
- Fix bug with rescheduling jobs to a past time


app/views/shared/_awards.html.rb
```diff
@@ -84,19 +84,7 @@ def content # rubocop:todo Metrics/PerceivedComplexity
...
-    elsif project.token._token_type_qtum?
-      image_tag 'ledger.png', alt: 'Ledger'
-    elsif project.token._token_type_ada? || project.token._token_type_btc?
-      image_tag 'trezor.png', alt: 'Trezor'
-    elsif project.token._token_type_eos?
-      image_tag 'eos.png', alt: 'Eos'
-    elsif project.token._token_type_xtz?
-      image_tag 'tezos.png', alt: 'Tezos'
-    end
+    image_tag 'metamask2.png', alt: 'Metamask2'
```
Alexey1100, line 87: This is legacy Awards page which is not used anymore, but keeps triggering SimpleCov.
We might want to delete it from the repo.

app/jobs/ore_id_opt_in_sync_job.rb
```diff
@@ -23,6 +23,14 @@ def perform(id)
   end
 
   def reschedule(ore_id)
-    self.class.set(wait: ore_id.next_sync_allowed_after(scale: :linear) - Time.current).perform_later(ore_id.id)
+    self.class.set(wait: wait_to_perform(ore_id)).perform_later(ore_id.id)
```
aquabu, line 26: Seems like the `reschedule` and `wait_to_perform` code is repeated in several jobs. Could it be factored out into a shared module?

app/views/shared/_awards.html.rb
aquabu: Yes, if we're not using it we should delete it and tag the last commit where it's present so we can reference later if needed.

aquabu (approved):

PR #1772, opened 2020-12-16
Add Algorand Security Token transactions support

Alexey1100:
This PR includes:

- Implement wrappers for validating and creating Algorand Security Token transactions
- Create transactions in three formats (Algosinger, Algorand Indexer, ChainJS)
- Sign Algorand Security Token transactions with OreId
- Validate transactions using Algorand Indexer
- Integrate wrappers into `BlockchainTransaction#on_chain` methods
- Add `fallback_state` stored in session to recover from failures during OreId browser interactions and cancel a transaction if present

Resolves:
https://www.pivotaltracker.com/story/show/175049231
https://www.pivotaltracker.com/story/show/175049220


app/lib/comakery/algorand/tx/app/security_token/transfer_group_lock.rb
```diff
@@ -0,0 +1,9 @@
+class Comakery::Algorand::Tx::App::SecurityToken::TransferGroupLock < Comakery::Algorand::Tx::App
+  def valid_app_args?(blockchain_transaction)
+    transaction_app_args[0] == 'transfer group'
+    transaction_app_args[1] == 'lock'
+    transaction_app_args[2] == 'from group id' # TODO: Update to use actual value
+    transaction_app_args[3] == 'to group id' # TODO: Update to use actual value
+    transaction_app_args[4] == Time.current.to_i.to_s # TODO: Update to use actual value
+  end
+nd
```
aquabu, line 9: needs an `e`

app/lib/comakery/algorand/tx/app.rb
```diff
@@ -7,21 +7,91 @@ def initialize(blockchain, hash, app_id)
     @app_id = app_id
   end
 
+  def to_object(blockchain_transaction)
```
Alexey1100, line 10: Thinking about providing `blockchain_transaction` instead of `hash` during initialisation to not pass it around here and in validation calls.
Will require refactoring `Comakery::Algorand::Tx` and `Comakery::Erc20::Tx` along with their calls.

app/lib/comakery/algorand/tx.rb
```diff
@@ -43,27 +53,27 @@ def confirmed?(_number_of_confirmations = 1)
     confirmed_round.present? && current_round >= confirmed_round
   end
 
-  def valid?(blockchain_transaction)
+  def valid?(_)
```
Alexey1100, line 56: Keep parameter for consistency with Comakery::Eth::Contract::Erc20::Tx

app/lib/comakery/algorand/tx/app.rb
```diff
@@ -1,27 +1,90 @@
...
-  # amount 100 mean 1.00 when App decimal is 2
-  def amount
-    0 # TODO: Fix me
+  def valid_transaction_on_completion?
+    transaction_on_completion == app_transaction_on_completion
   end
 
-  def valid?(blockchain_transaction)
-    super && app_id.to_i == transaction_app_id
+  def valid?(_)
```
Alexey1100, line 83: Keep parameter for consistency with Comakery::Eth::Contract::Erc20::Tx

app/lib/comakery/algorand/tx.rb
```diff
@@ -43,27 +52,25 @@ def confirmed?(_number_of_confirmations = 1)
     confirmed_round.present? && current_round >= confirmed_round
   end
 
-  def valid?(blockchain_transaction)
+  def valid?(_ = nil)
```
aquabu, line 55: What is this argument used for?

app/lib/comakery/algorand/tx/app.rb
```diff
@@ -1,27 +1,140 @@
...
+        '0x' + a.unpack1('H*')
+      when Integer
+        '0x' + a.to_s(16)
+      else
+        raise "Unsupported app argument: #{a}:#{a.class}"
+      end
+    end
+  end
+
+  # Note: Algosinger appArgs format
```
aquabu, line 46: Algosinger  -> Algosigner

app/lib/comakery/algorand/tx.rb
Alexey1100: @aquabu 

It's left for interface compatibility with `Comakery::Eth::Contract::Erc20::Tx` subclasses (which accept the blockchain transaction during validation rather than initialisation), since `BlockchainTransaction` doesn't know the type when calling the validation (see `BlockchainTransaction#valid_on_chain?`)

app/lib/comakery/algorand/tx/app/security_token/burn.rb
```diff
@@ -0,0 +1,22 @@
+class Comakery::Algorand::Tx::App::SecurityToken::Burn < Comakery::Algorand::Tx::App
+  def app_args
```
aquabu, line 2: Can these app functions be consolidated into one app class? One class per method seems like it obscures the total app behavior.

app/lib/comakery/algorand/tx/app/security_token/set_address_permissions.rb
```diff
@@ -0,0 +1,17 @@
+class Comakery::Algorand::Tx::App::SecurityToken::SetAddressPermissions < Comakery::Algorand::Tx::App
+  def app_args
+    [
+      'setAddressPermissions',
+      (blockchain_transaction.blockchain_transactable.account_frozen ? 1 : 0),
```
aquabu, line 5: Non blocking comment. There are a lot of chained methods/associations here. Maybe we should consider refactoring... see the Law of Demeter. For example: https://medium.com/better-programming/demeters-law-don-t-talk-to-strangers-87bb4af11694

app/lib/comakery/algorand/tx/asset.rb
```diff
@@ -21,7 +32,7 @@ def amount
     transaction_data.dig('asset-transfer-transaction', 'amount') || 0
   end
 
-  def valid?(blockchain_transaction)
-    super && asset_id.to_i == transaction_asset_id
+  def valid?(_ = nil)
```
aquabu, line 35: Seems like we are just throwing away any argument here? What's the purpose? Is this because we have to take one argument to implement a `valid?(arg)` interface

app/lib/comakery/algorand/tx/app/security_token/burn.rb
Alexey1100: @aquabu each subclass represents a **transaction type** based on action (transfer, asset_transfer, asset_opt_in, app_opt_in, app_transfer, etc.) and inherits from parent type – Burn in this case inherits from a generic app transaction (only having custom app args and app accounts), which itself inherits from Algorand transaction (having custom transaction fields). 
 
Since the purpose of each instance is to validate and generate the transaction (the logic which depends only on the type) I think it's totally make sense to build the hierarchy based on that.

app/lib/comakery/algorand/tx/app/security_token/set_address_permissions.rb
Alexey1100: I agree, the relationships between BlockchainTransaction, Blockchain and Tx wrapper classes are already asking for refactoring for a while :)

I think we can do it once we finish the Algorand Integration, as we initially planned, so we will have at least two different security tokens implemented.

app/models/blockchain_transaction.rb
```diff
@@ -101,8 +101,8 @@ def contract
 
   private
 
-    def populate_data # rubocop:todo Metrics/CyclomaticComplexity
-      self.token ||= blockchain_transactable.token
+    def populate_data # rubocop:todo Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      self.token ||= blockchain_transactable.is_a?(Token) ? blockchain_transactable : blockchain_transactable.token
```
aquabu, line 105: Seems like we should make the blockchain transactable subclasses polymorphic so we don't have to implement any conditionals based on class here. The if statement should be encapsulated in the object and the objects should have a standard interface. By passing in the object and calling the standard method implementation of the interface it will return the right value.

By doing this it will also lower the cyclomatic complexity of the function.

I was trying to find a succinct example to illustrate. Sand Metz presents a lot of good material on this. Here's a succinct ok explanation:
https://cek.io/blog/2014/05/02/poodr-ch-5/

It's ok to do this in a follow up PR but it should be done.

app/models/blockchain_transaction_account_token_record.rb
```diff
@@ -4,7 +4,19 @@ def update_transactable_status
   end
 
   def on_chain
-    @on_chain ||= Comakery::Eth::Tx::Erc20::SecurityToken::SetAddressPermissions.new(token.blockchain.explorer_api_host, tx_hash)
+    @on_chain ||= if token._token_type_on_ethereum?
+      on_chain_eth
```
aquabu, line 8: Similar to the last comment, this should be handled with polymorphism instead of a conditional statement that switches on type.

app/models/blockchain_transaction_award.rb
```diff
@@ -5,13 +5,17 @@ def update_transactable_status
     blockchain_transactable.update!(status: :paid)
   end
 
-  def on_chain
+  def on_chain # rubocop:todo Metrics/CyclomaticComplexity
```
aquabu, line 8: Same. This should use polymorphism instead of switching on type.

aquabu (changes requested):

aquabu:
@Alexey1100 I made some significant refactoring recommendations. I know this PR is critical so they can be done as fast follow ons. But please consider how to refactor into polymorphic classes instead of switching on type. Also, it seems that the app functions should be in one class instead of many (unless I'm missing something important).

app/lib/comakery/algorand/tx/asset.rb
Alexey1100: Yep, pls see previous response for details.

app/models/blockchain_transaction.rb
Alexey1100: blockchain_transactable is already polymorphic, I think the exception here is that some of the blockchain transactables  belong to token, while some of them are token themselves.

We can avoid this particular condition by adding a token method to token itself, for the interface consistency.

app/models/blockchain_transaction_account_token_record.rb
Alexey1100: 🤔
While all tx subclasses provide the same interface, we still have to decide somewhere what subclass to load based on Blockchain and BlockchainTransactable properties.

Let me think how can we get rid of the condition, since creating a polymorphic `BlockchainTransaction*` model for each blockchain and transaction type combo seems a bit overkill.

Alexey1100:
@aquabu sure, I'll look into ways how to improve the associations between Model and wrappers, there's a couple of optimisations I've mentioned above.

> app functions should be in one class instead of many (unless I'm missing something important).
In our case class represents not an app but a transaction, which needs to be parsed and associated with a record in database.

app/models/blockchain_transaction_account_token_record.rb
aquabu: Naively speaking the logic would be replaced with `token.on_chain`

Instead of `@on_chain = # conditional statement based on type of token`

Alexey1100: Ohh right, we can partially move the tx wrapper initialiser to the `TokenType#on_chain` and then `BlockchainTransaction` subclass can decide the tx type based on it's own type. 

E.g:

```ruby
# token_type/algorand_security_token.rb
def on_chain_prefix
  "Comakery::Algorand::Tx::App::SecurityToken"
end

# blockchain_transaction_account_token_record.rb
def on_chain_name
  "SetAddressPermissions"
end

# blockchain_transaction.rb
def on_chain
  (token.token_type.on_chain_prefix + on_chain_name).constantize.new(self)
end
```

app/models/blockchain_transaction.rb
aquabu: Yes, I think it would be beneficial to encapsulate that. As a user of the object it's preferable to not need to know that detail and write special cases for it.

app/models/blockchain_transaction_account_token_record.rb
Alexey1100: @noah we base the condition here on 3 factors:

1) token
2) blockchain_transaction type
3) transfer type, in case if blockchain_transaction type is transfer

app/lib/comakery/algorand/tx.rb
aquabu: ok. it's probably worth a comment

app/controllers/concerns/ore_id_callbacks.rb
```diff
@@ -24,10 +24,20 @@ def crypt
   end
 
   def state(**additional_params)
-    @state ||= crypt.encrypt_and_sign({
+    @state = crypt.encrypt_and_sign({
       account_id: current_account.id,
       redirect_back_to: params[:redirect_back_to] || request.referer
     }.merge(additional_params).to_json)
+
+    session[:sign_ore_id_fallback_state] = @state
```
Riddlerrr, line 32: 👍

Riddlerrr (approved):
Great work!

Alexey1100:
It's a bit hard to manually test the functionality right now, given that we don't allow adding Algorand wallets manually and wallet opt-in view is broken right now.

I'll provide some steps today.

aquabu (approved):

Alexey1100:
Ok, this is how we can test the whole set up right now 😅
We need to set up two accounts, one of them having a set of roles given externally with Algosigner. 

### Contributor account

1. Link OREID
2. Create app `optIn` tx for OREID wallet address **(from Wallets page)**

### Project admin account

1. Link OREID
2. Create app `optIn` tx for OREID wallet address **(from Wallets page)**
3. Create app `setRole` tx for OREID wallet address to give wallet role `7` (Reserve + Transfer Rules + Wallets) **(using Algosigner with contract admin account)**
4. Update admin account reg group on Accounts page
5. Update contributor account reg group on Accounts page
6. Create mint transfer to admin account 
7. Create transfer from admin to contributor account 

#### Creating `setRole` tx with Algosigner:
1. Install the [Algosigner](https://www.purestake.com/technology/algosigner/)
2. Set network to AlgorandTest and import account with contract admin role
2. Go to https://purestake.github.io/algosigner-dapp-example/tx-test/signTesting.html
3. Uncheck `Use Default Tx`, Check `Send Transaction`
4. Click `Pay`
5. Paste following transaction object and submit the tx:

```json
{"type":"appl","from":"IKSNMZFYNMXFBWB2JCPMEC4HT7UECC354UDCKNHQTNKF7WQG3UQW7YZHWA","fee":1000,"firstRound":12141970,"lastRound":12142970,"genesisID":"testnet-v1.0","genesisHash":"SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=","appIndex":13997710, "appOnComplete":0, "appArgs": ["Z3JhbnRSb2xlcw==", [7]],"appAccounts": ["IKSNMZFYNMXFBWB2JCPMEC4HT7UECC354UDCKNHQTNKF7WQG3UQW7YZHWA"]}
```

Where:
- `firstRound` and `lastRound` should be less and greater than current blockchain round accordingly
- `from` should be contract admin account imported to Algosigner
- `appAccounts` should contain address of the account we want grant the role to

PR #1774, opened 2020-12-17
ORE ID: Introduce provisioning of multiple tokens

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/176078030

This PR moved provision logic from `OreIdAccount` into `WalletProvision`. This way we can provision as many tokens as we want.

This PR didn't include data migration and removing data from `OreIdAccount#provision_stage`.


app/controllers/api/v1/wallets_controller.rb
```diff
@@ -69,17 +70,59 @@ def wallet_params
...
-      r.delete(:provision)
       r
     end
 
     def redirect_url
       @redirect_url ||= params.fetch(:body, {}).fetch(:data, {}).fetch(:redirect_url, nil)
     end
+
+    def tokens_to_provision
+      @tokens_to_provision ||=
```
aquabu, line 86: could refactor this section to be simpler

app/models/token_type.rb
```diff
@@ -17,4 +17,8 @@ def self.list
   def self.append_to_list(token_type)
     list.merge(token_type => (list.values.max + 1))
   end
+
+  def can_be_provisioned?
```
aquabu, line 21: should add this method to the template as well.

aquabu (approved):

PR #1775, opened 2020-12-18
Updated heroku deployment

aquabu:
Revised version of improved deployment scripts that is rebased against latest commits from the acceptance branch.

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -39,6 +39,8 @@
 PreCommit:
   BundleCheck:
     enabled: true
+  MergeConflicts:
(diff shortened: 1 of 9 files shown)


Riddlerrr:
image

I figured out the reason of failing of initial deploy.
Heroku created DB before we run `db:prepare` so the command run all migrations because BD already exists.

I've added separate "smart" `bin/after_deploy_script` which load db schema only if needed.

P.S. @aquabu Could you remove all 3 deploytest apps from heroku. I don't have access to do it.

PR #1782, opened 2020-12-28
Issue#1779 Migrate to ActiveStorage

Lerek:


app/controllers/accounts_controller.rb
```diff
@@ -231,7 +231,14 @@ def project_decorate(project, award = nil)
 
     def account_decorate(account)
       account.as_json(only: %i[email first_name last_name nickname date_of_birth country ethereum_auth_address]).merge(
-        image_url: account.image.present? ? Refile.attachment_url(account, :image, :fill, 190, 190) : nil
+        image_url: account_image_url(account)
```
aquabu, line 234: This isn't blocking for this commit. But we should probably move the controller's decorator functions into the corresponding decorator classes since we have established that pattern.

app/controllers/missions_controller.rb
```diff
@@ -194,21 +196,31 @@ def set_form_props
...
-        whitelabel_favicon_url: @mission&.whitelabel_favicon&.present? ? Refile.attachment_url(@mission, :whitelabel_favicon, :fill, 64, 64) : nil
+        logo_url: mission_image_path(@mission.logo, 800, 800),
+        image_url: mission_image_path(@mission.logo, 1200, 800),
+        whitelabel_logo_url: mission_image_path(@mission.logo, 1000, 200),
+        whitelabel_logo_dark_url: mission_image_path(@mission.logo, 1000, 200),
+        whitelabel_favicon_url: mission_image_path(@mission.logo, 64, 64)
       }
     end
 
+    def mission_image_path(image, width, height)
```
aquabu, line 209: maybe this should go in the mission decorator?

app/controllers/pages_controller.rb
```diff
@@ -59,8 +59,15 @@ def featured_mission_props(mission)
...
       mission.as_json(only: %i[id name]).merge(
         mission_url: mission_url(mission),
-        image_url: mission.image.present? ? Refile.attachment_url(mission, :image, :fill, 231, 231) : nil,
+        image_url: mission_image_path(mission, 231),
         projects_count: mission.projects.public_listed.active.count
       )
     end
+
+    def mission_image_path(mission, size)
+      GetImageVariantPath.call(
```
aquabu, line 68: it does look like we could remove duplication for image methods in the controller by moving this to the mission decorator class.

app/interactors/get_image_path.rb
```diff
@@ -0,0 +1,6 @@
+class GetImagePath
```
aquabu, line 1: please add a comment as to when to use this vs when to use GetImagePath vs GetImageVariantPath.

app/interactors/get_image_variant_path.rb
```diff
@@ -0,0 +1,7 @@
+class GetImageVariantPath
```
aquabu, line 1: please add a comment as to when to use this vs when to use GetImagePath vs GetImageVariantPath.

app/interactors/attachment/get_path.rb
```diff
@@ -0,0 +1,14 @@
+module Attachment
```
aquabu, line 1: For the interactors please add a comment about when to use them. It seems that some of the interactors are not intended to be called directly because they are meant to be part of other interactors that organize them. Good to document the intended use.

aquabu (approved):
@Lerek nice PR. I suggested a minor refactoring and requested a few comments for interactors. I'm approving since these can be addressed later and shouldn't block the PR.

Lerek:
> @Lerek nice PR. I suggested a minor refactoring and requested a few comments for interactors. I'm approving since these can be addressed later and shouldn't block the PR.

@aquabu thank you for the review!

I've added comments for newly introduced interactors. Basically, every of them can be used by it's own, only `Attachment::SetFallbackPath` makes no sense to be solo, as it's intention was DRYing fallback logic.

As for other comments -> i completely agree, but such refactoring deserves several separated PRs :) Didn't want to put too much effort on experimental idea, wich might not work out at final

aquabu:
@Lerek great thanks. Makes sense.

An upstream merge updated a gem, so there's a merge conflict to resolve. Other than that this looks good to me.

Issue #1777, opened 2020-12-24
If slack or discord environment variables are not present do not show slack or discord login buttons on the signup page


aquabu:
# Goal

We want to make deployment of new CoMakery Server instances to Heroku as easy as possible. Slack and Discord require some configuration for each new domain and this can slow down initial server deployment. For this reason we don't want to require Slack and Discord configuration environment variables. Accordingly, if the initial configuration does not have the Discord or Slack configuration variables we don't want to display the Slack and Discord logins. 

## Background

There will be many deployments of the comakery server. Some server instances will be deployed with the whitelabel layout and functionality like https://whitelabel-audit.comakery.com . Others will be deployed with the full comakery social features like www.comakery.com.

### Example Current Whitelabel Pages

* Sign Up https://whitelabel-audit.comakery.com/accounts/new
* Sign In https://whitelabel-audit.comakery.com/session/new

### Example Current Non-Whitelabel (Default) Pages

* Sign Up https://www.comakery.com/accounts/new
* Sign In https://www.comakery.com/session/new

# Desired Changes On The Non-Whitelabel (Default) Instance of the Server

- [ ] On the **signup page** of the **non-whitelabel** version of the product if the DISCORD_BOT_TOKEN or DISCORD_CLIENT_ID environment variables are not set, **do not display the Discord login button**
- [ ] On the **login page** of the **non-whitelabel** version of the product if the DISCORD_BOT_TOKEN or DISCORD_CLIENT_ID environment variables are not set, **do not display the Discord login button**
- [ ] On the **signup page** of the **non-whitelabel** version of the product if the SLACK_API_KEY or SLACK_API_SECRET environment variables are not set, **do not display the Slack login button**
- [ ] On the **login page** of the **non-whitelabel** version of the product if the SLACK_API_KEY or SLACK_API_SECRET environment variables are not set, **do not display the Slack login button**
- [ ] Make the slack and discord fields optional in the app.json file with`"required": false` - so that new deployments to heroku with the "Deploy To Heroku" button in the README do not require the Slack and Discord environment variables

# Login Configuration That Stays The Same

### For now the whitelabel sign in page will continue to look like this:

<img width="1004" alt="Screen Shot 2020-12-23 at 8 07 33 PM" src="https://user-images.githubusercontent.com/7595/103059641-88d58580-455a-11eb-9398-c4ce2e81a877.png">

### And the white label sign up page will continue to look like this

[...]

PR #1783, opened 2020-12-28
issue#1777 turn off discord/slack if not configured

matjacdev515:
#1777  enable/disable slack or discord buttons on signup/login according to env variables.


Riddlerrr (dismissed):
The implementation looks good 👍 Well done!

But we have to add a few tests to cover the changes:
image

Alexey1100:
Yep looks good :)

We will also need to delete these variables from `.env.required` file, since we have some logic around that (see `env_verifier.rb`)

Alexey1100:
Also I believe, we have a Channels section in project settings (`/projects/:id/edit`) which tells you to sign in with Slack/Discord to be able to use it.
I think we want to adjust that as well.

Alexey1100 (approved):
👍

aquabu (approved):

aquabu:
@Riddlerrr - @matjacdev515 fixed the test coverage issue you pointed out. I'm approving for merge.

Riddlerrr (approved):
👍

Issue #1778, opened 2020-12-24
Delete old migrations in a safe way for old and new deployments


aquabu:
# Goal

We have a lot of old migrations. If you were to `rake db:create` a database and then run `rake db:migrate` the migrations would fail because some of the methods used in migrations are no longer available. This can of course be avoided by running `rake db:prepare`. However, we should clean up the old migrations since they are no longer useful.

We also want to verify that the migrations run in the future and don't break by including a migration test during continuous integration (CI) which runs as a Github Action for the Repository. CI output is visible here: https://github.com/CoMakery/comakery-server/actions

We also want to be able to review the old migrations in the future even though we don't want to keep maintaining them. So we should tag the commit where they are removed.

## Requirements Checklist

A solution to this should satisfy the following requirements
- [ ] Add a tag to find an archive of old migrations `archive-migrations-to-[schema]`
- [x] remove the old migrations before June 1 2020
- [x] keep the migrations from the second half of 2020 since some of these may not have been run in all deployments
- [x] running `rake db:create` and `rake db:migrate` should migrate up to the current schema successfully without failures
- [ ] running migrations on existing staging, test and production environments should not cause problems with their schemas
- [ ] the GitHub workflow on CI should run all remaining migrations up and migrations down successfully as scripted in https://github.com/CoMakery/comakery-server/blob/acceptance/.github/workflows/test.yml

ravi4rails:
@aquabu can you please confirm the second item in checklist?

aquabu:
@ravi4rails yes, the idea is to shorten the number of migrations that will run. There may be some instances of CoMakery that need to run older migrations. All of the CoMakery instances we manage are up to date but there may be some versions deployed by other hosts that have a different version.

ravi4rails:
@aquabu I pushed changes. And this worked using the reference you shared:
https://github.com/jalkoby/squasher

PR #1784, opened 2020-12-28
Remove old migrations

ravi4rails:
#1778


db/migrate/20200528155545_init_schema.rb
```diff
@@ -0,0 +1,469 @@
+class InitSchema < ActiveRecord::Migration[4.2]
```
aquabu, line 1: migration version may need to be 6.0

app/models/account.rb
```diff
@@ -1,7 +1,7 @@
 # Allow usage of has_and_belongs_to_many to avoid creating a separate model for accounts_projects join table:
 # rubocop:disable Rails/HasAndBelongsToMany
 
-class Account < ApplicationRecord
```
Riddlerrr, line 4: Why did you change it? Since Rails 5`ApplicationRecord ` is parent  class for models

db/migrate/20200528155545_init_schema.rb
ravi4rails: @aquabu done

app/models/account.rb
ravi4rails: @Riddlerrr I changed this to `ActiveRecord::Base` because after deleting migrations I was getting this error
`TypeError: superclass mismatch for class <ModelName>` for account, authentication and project models. 

and after replacing `ApplicationRecord` with `ActiveRecord::Base` this was solved.

ravi4rails: But I just checked again by `ApplicationRecord` as parent class and no errors now. May be this was because I forget to add 
6.0 here `class InitSchema < ActiveRecord::Migration[4.2]`  .

I am pushing this quickly

Riddlerrr: 👍

db/migrate/20170116023520_write_new_award_column_values.rb
```diff
@@ -1,7 +1,6 @@
 class WriteNewAwardColumnValues < ActiveRecord::Migration[4.2]
```
aquabu, line 1: It seems like this migration should have been deleted.

.github/workflows/test.yml
```diff
@@ -135,7 +135,7 @@ jobs:
       
     - name: Precompile assets
       run: bundle exec rails assets:precompile
-      
+
```
aquabu, line 138: Need to test migrations up with `rake db:migrate` and migration down with `rake db:migrate:down VERSION=`, 

Then test the more common way to create the database for a new install by dropping the db and running `rake db:prepare`

aquabu (changes requested):
A few requested changes:
* there seems to be one earlier migration that should be deleted
* need CI testing code to test migrations

These are described in the comments.

Gemfile
```diff
@@ -154,3 +154,5 @@ gem 'validate_url', '~> 1.0'
 gem 'overcommit', '~> 0.53.0'
 
 gem 'fuubar', '~> 2.5'
+
+gem 'squasher'
```
Riddlerrr, line 158: I think we don't need to include this gem into `Gemfile` because it's one-time usage gem and we don't want to maintain/update it.

ravi4rails: Yes @Riddlerrr I not needed this anymore

ravi4rails: removed and pushed

bin/squasher
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
```
Riddlerrr, line 1: @ravi4rails Don't forget to remove this file too.

.github/workflows/test.yml
ravi4rails: yes got this. Just one thing this should be in current branch's test.yml or at acceptance? For now I am doing with current branch's test.yml

bin/squasher
ravi4rails: yup, removed this one too.

db/migrate/20170116023520_write_new_award_column_values.rb
ravi4rails: just kept this because removing this migration will have a merge conflict with acceptance. But added again, I think I need to discuss this you on call.

.github/workflows/test.yml
ravi4rails: regarding this command: `rake db:migrate:down VERSION=` any specific version for this?

Riddlerrr: > regarding this command: rake db:migrate:down VERSION= any specific version for this?

This command migrates down only one specific migration which you set in `VERSION`.
So if we want to migrate down all existing migrations better to use `rake db:rollback STEP=n` where `n` is the count of all migrations.

You can find `n` by run the command: `bin/rails db:migrate:status | grep -c 'up'`
↣ zsh  sergey@MacBook-Pro-Andronov local  2021-01-12 09-16-39

aquabu: ok 👍

Riddlerrr:
@ravi4rails I've updated with acceptance branch to avoid conflict.
Also, I've fixed CI tests so I think it's ready to merge.

cc @aquabu

.github/workflows/test.yml
Riddlerrr: I think we don't need to test `down` migrations because some of them can be irreversible.

aquabu (approved):

PR #1785, opened 2020-12-29
[#1776] Change account build profile page 

### LINKED ISSUE #1776: Change account build profile page form fields shown after signup
When a user signs up they are currently presented with this form on the /accounts/build_profile page:
Screen Shot 2020-12-23 at 5 31 40 PM

The fields should be changed to:
Screen Shot 2020-12-23 at 5 44 14 PM

All four of the fields should be required including:
* First Name
* Last Name
* Date of Birth
* Country of Citizenship

The signup process should include an integration test.

swimsoft:
**Resolves #1776** 

Updated account build profile page form and style

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -72,7 +72,7 @@ class Account < ApplicationRecord
   validates :email, presence: true, uniqueness: { scope: %i[managed_mission], case_sensitive: false }
   # rubocop:enable Rails/UniqueValidationWithoutIndex
   validates :password, length: { minimum: 8 }, if: :password_required
-  validates :first_name, :last_name, :country, :specialty, presence: true, if: :name_required
+  validates :first_name, :last_name, :country, presence: true, if: :name_required
   validates :date_of_birth, presence: { message: 'should be present in correct format' }, if: :name_required
   validates :nickname, uniqueness: true, if: -> { nickname.present? }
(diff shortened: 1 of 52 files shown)


app/models/account.rb
```diff
@@ -72,7 +72,7 @@ class Account < ApplicationRecord
   validates :email, presence: true, uniqueness: { scope: %i[managed_mission], case_sensitive: false }
   # rubocop:enable Rails/UniqueValidationWithoutIndex
   validates :password, length: { minimum: 8 }, if: :password_required
-  validates :first_name, :last_name, :country, :specialty, presence: true, if: :name_required
+  validates :first_name, :last_name, :country, presence: true, if: :name_required
```
Alexey1100, line 75: Wondering if missing specialty for newly created profiles would break contributor experience calculations for tasks.
I think, it needs additional investigation here.

Cc: @aquabu

Alexey1100 (commented):
Otherwise looks good :)

app/models/account.rb
Alexey1100: Just checked and I think we are safe.
Since looks like we are only using account specialty for informational purposes.

When the task is completed we are creating or incrementing account experience based on award's specialty, not account's one.

Alexey1100 (approved):

PR #1799, opened 2021-01-06
Fix ActiveStorage population data migrations

Lerek:

PR #1800, opened 2021-01-06
Rubocop: Ignore data_migrations files

Riddlerrr:
In #1799 I noticed that we checking data migration files with robocop. I think it's not needed since we ignore regular migrations.


Alexey1100:
🤔
It might help us to catch some pitfalls, since code in the data migrations deals with regular ActiveRecord calls.

Riddlerrr:
Yes, I have mixed feelings too. 50/50
Ok, let's leave it as is and close the PR

PR #1804, opened 2021-01-07
API client can set extended params when creating the wallet

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/176217650

Docs update:
image


app/services/wallet_creator.rb
```diff
@@ -1,71 +1,116 @@
...
+    def sanitize_params(params)
+      return [] if params.blank?
 
-      if wrong_tokens.any?
-        wallet.errors.add(:tokens_to_provision, "Some tokens can't be provisioned: #{wrong_tokens}")
-        return false
-      end
+      error = { error: 'Wrong format. It must be an Array.' }
+      return error unless params.is_a?(Array)
+      return error unless params.first.is_a?(ActionController::Parameters)
```
matjacdev515, line 87: Do we really care about this? Think regular Hash will also work

app/services/wallet_creator.rb
```diff
@@ -1,71 +1,116 @@
...
-      true
+      params
     end
 
-    def build_wallet_provisions(wallet, tokens_to_provision)
-      tokens_to_provision.each do |token_to_provision|
-        wallet.wallet_provisions.new(token_id: token_to_provision)
+    def validate_params_type
+      wallet.errors.add(:tokens_to_provision, params[:error]) if params.is_a?(Hash) && params.key?(:error)
+      wallet.errors.empty?
```
matjacdev515, line 94: Can we do this check only once inside of `should_be_provisioned?`? These methods will only add errors.

app/services/wallet_creator.rb
```diff
@@ -1,71 +1,116 @@
...
+    end
+
+    def validate_token_id_provided
+      wallet.errors.add(:tokens_to_provision, 'token_id param must be provided') unless params.all? { |p| p.key?(:token_id) }
+      wallet.errors.empty?
+    end
+
+    def validate_tokens_availability
+      unavailable_token_ids = params.map { |t| t.fetch(:token_id) }.map(&:to_i) - available_tokens_to_provision.pluck(:id)
```
matjacdev515, line 103: `params.map { |t| t.fetch(:token_id).to_i }` this will reduce the number of iterations

aquabu:
@Riddlerrr could you address comments from @matjacdev515 and then get him to re-review?

app/models/wallet.rb
```diff
@@ -7,6 +7,7 @@ class Wallet < ApplicationRecord
   has_many :token_opt_ins, dependent: :destroy
   has_many :wallet_provisions, dependent: :destroy
   has_many :awards, foreign_key: :recipient_wallet_id, inverse_of: :recipient_wallet, dependent: :nullify
+  has_many :account_token_records, dependent: :destroy
```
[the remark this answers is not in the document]
Riddlerrr: Why `:destroy` not work?

app/services/wallet_creator.rb
Riddlerrr: Yeah, agree. Initially, I wanted to check using strong params but in the final implementation it's not necessary.

app/services/wallet_creator.rb
Riddlerrr: I've refactored validation flow

app/services/wallet_creator.rb
```diff
@@ -1,71 +1,116 @@
...
     end
 
     account.save if errors.empty?
-
     [wallets, errors]
   end
+end
 
-  private
+class WalletCreator::Provision
```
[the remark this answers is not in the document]
Riddlerrr: We can but I don't see reasons to do this. This helper class needs only for `WalletCreator`, so why not to keep it in the same class. I find it more readable because you no have to jump to another file.

app/models/wallet.rb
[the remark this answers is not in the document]
matjacdev515: `destroy` fetches all objects for DB, runs callbacks on them then does a separate DELETE SQL query per object.

 `delete_all ` just does a single `DELETE ... WHERE id IN (....)` SQL query for whole collection .  
 
 Since there is no any `on: :destroy` callbacks for `account_token_records`, from performance point of view `delete_all` is much preferred.

app/services/wallet_creator.rb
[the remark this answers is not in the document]
matjacdev515: At leat Rails 6 Zeitwerk autoloader wouldn't work in that case

https://www.rubyguides.com/2019/08/autoloading-in-ruby/

> Define one class / module per file

And this is a good practice

app/models/wallet.rb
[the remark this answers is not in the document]
Riddlerrr: We definitely won't get a problem with performance in this case and we use `depends: :destroy` everywhere in the project, so I want to be consistent and use it `:destroy` here.

app/services/wallet_creator.rb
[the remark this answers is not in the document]
Riddlerrr: Good point 👍 Okay, I'll move to separate file

app/models/wallet.rb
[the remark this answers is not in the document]
matjacdev515: It is not about consistency. If we can avoid N+1 queries and additional load in memory by this simple change why we shouldn't do this?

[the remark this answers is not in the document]
aquabu: let's resolve this in a subsequent task for the whole project.

app/controllers/api/v1/wallets_controller.rb
```diff
@@ -78,7 +78,10 @@ def wallets_params
       wallets = params.fetch(:body, {}).fetch(:data, {}).require(:wallets)
 
       wallets.map do |wallet_params|
-        wallet_params.permit(:blockchain, :address, :source, :tokens_to_provision, :name)
+        wallet_params.permit(
+          :blockchain, :address, :source, :name,
+          tokens_to_provision: %i[token_id max_balance lockup_until reg_group_id account_id account_frozen]
```
Alexey1100, line 83: Looks like `account_id ` is not needed here

app/controllers/api/v1/wallets_controller.rb
```diff
@@ -78,7 +78,10 @@ def wallets_params
       wallets = params.fetch(:body, {}).fetch(:data, {}).require(:wallets)
```
Alexey1100, line 78: Should the parameter be named `wallet`?

app/controllers/api/v1/wallets_controller.rb
Riddlerrr: Right 👍

Riddlerrr: Fixed here https://github.com/CoMakery/comakery-server/pull/1832

PR #1806, opened 2021-01-08
API can create multiple wallets

matjacdev515:
Task: 
https://www.pivotaltracker.com/story/show/176238543
https://www.pivotaltracker.com/story/show/176338450


app/controllers/api/v1/wallets_controller.rb
```diff
@@ -11,14 +11,42 @@ def index
...
-      render 'show.json', status: :created
-    else
-      @errors = @wallet.errors
+      if @wallets.all?(&:persisted?)
+        render 'index.json', status: :created
+      else
+        @errors = {}
 
-      render 'api/v1/error.json', status: :bad_request
+        @wallets.each_with_index.map do |wallet, i|
```
Riddlerrr, line 22:
```suggestion
        @wallets.map.with_index do |wallet, i|
```

Riddlerrr: It just reads a little bit better.

app/models/wallet.rb
```diff
@@ -31,4 +34,19 @@ def coin_balance
...
     balance
   end
+
+  def set_primary_flag
+    self.primary_wallet = !Wallet.exists?(account_id: account_id,
+                                          _blockchain: _blockchain,
+                                          primary_wallet: true)
+  end
+
+  def mark_another_wallet_as_primary
```
Riddlerrr, line 44: Why another? We mark `first` wallet as primary here. Would it better name `mark_first_wallet_as_primary `?

Riddlerrr (approved):
Well done 👍

Riddlerrr:
Oh, as I see there is new PR with the same changes: https://github.com/CoMakery/comakery-server/pull/1809
So I close this PR

PR #1808, opened 2021-01-10
added whitelable env to app json

ravi4rails:


ravi4rails:
@Alexey1100 need your assistance in this.

Alexey1100:
I think we decided to add an initialiser which will check if whitelabel env variable is set to true.
If so, it creates a whitelabel mission (see mom.rb:667) with domain name extracted from `APP_HOST` variable.
This way all further requests to the instance would be routed to that whitelabel mission.

We might want to add a note into Readme on configuring mission details in rails console, such as name, email, logo, favicon and api keys.
Alternatively we can detect that whitelabel mission is unconfigured and create a simple welcome screen where first-time instance visitor can set all the details above.
(cc: @aquabu)

Let me know if you need any other info (:

Riddlerrr:
@Alexey1100 Great idea with a new welcome page, I support this 👍 
It should be a separate task of course.

ravi4rails:
@Alexey1100 you mean `mom.rb:617` right?

Alexey1100:
@ravi4rails mom:628 in this branch

ravi4rails:
ahh, yes got that @Alexey1100

aquabu:
@Alexey1100 definitely a good idea to have a welcome page for configuring the app. Given our priorities right now I would like to defer implementing the welcome page to later.

@ravi4rails For now if we can auto create a mission with the configuration variables on deployment that would be great.

To make that possible I think you will need to:
* Set the mission whitelabel field to "Enabled"
* If the APP_HOST is present, set the Whitelabel Domain to the APP_HOST value
* Make mission logo and mission image optional fields on the model and on the mission admin edit page. They are currently shown as required on `/missions/:id/edit
* You may need to use default white images of the correct dimensions to keep from breaking page layouts.

Below are some white images to use as default values for the mission. To see them you may need to switch to dark mode in GitHub > Account Settings > Appearance.

## Mission Logo 800px x 800px




## Mission Image 1200px x 800px

lib/tasks/whitelabel_mission.rake
```diff
@@ -0,0 +1,31 @@
+namespace :mission do
```
aquabu, line 1: @Alexey1100 @Riddlerrr would this default mission be better in db:seed?

I believe we are running `rake db:prepare` after deployment to Heroku with app.json and that task runs:
```
rake db:create 
rake db:seed  
rake db:migrate
```

alternatively, we could just run this by including the script in the [app.json](https://github.com/CoMakery/comakery-server/blob/6b03b88c298a38c18e2dce036bd2f524b3537766/app.json#L8) postdeploy array.

aquabu (changes requested):

lib/tasks/whitelabel_mission.rake
Riddlerrr: I think better to create the white mission using `postdeploy` in `app.json` file. It should be run only once this way.
Also, I think we can remove running `db:prepare` because we run this on each release: https://github.com/CoMakery/comakery-server/blob/27840d6c0cd428d328bbe8a2588984133a5c48f1/Procfile#L3

But it's should be checked, I not sure it will work on the very first deploy. If my suggestion is true we can just replace
`"postdeploy": "bundle exec rails db:prepare"` to `"postdeploy": "bundle exec rails mission: default_whitelabel_mission"`

ravi4rails: Yes did that. and this worked.

app.json
```diff
@@ -126,6 +126,18 @@
       "description": "Username for Sidekiq admin panel",
       "required": false
     },
+    "WHITELABEL": {
+      "description": "Deploy white lable layout to heroku.",
+      "required": true
+    },
+    "WHITELABEL_API_PUBLIC_KEY": {
+      "description": "Required for creating default whitelabel mission.",
+      "required": true
```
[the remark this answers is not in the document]
ravi4rails: done

app/controllers/application_controller.rb
```diff
@@ -247,6 +247,7 @@ def compare_all(*pairs)
     end
 
     def set_whitelabel_mission
+      puts current_domain
```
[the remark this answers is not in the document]
ravi4rails: done

PR #1809, opened 2021-01-10
Transfer API Wallet Selection

matjacdev515:
Task https://www.pivotaltracker.com/story/show/176238550


app/models/award.rb
```diff
@@ -30,6 +30,7 @@ class Award < ApplicationRecord
   belongs_to :issuer, class_name: 'Account', touch: true
   belongs_to :channel, optional: true
   belongs_to :specialty
+  belongs_to :wallet, optional: true
```
aquabu, line 33: There are two design decisions here:
* I think it's a good idea to denormalize the final recipient wallet address and store it as a string in the award table instead of an association. It may be that only occurs after the blockchain transaction payment is initiated and before then the user can change their primary wallet.
* at minimum the name of the association should indicate if the associated wallet is the the sender or receiver wallet

app/models/wallet.rb
```diff
@@ -31,4 +35,19 @@ def coin_balance
...
     balance
   end
+
+  def set_primary_flag
+    self.primary_wallet = !Wallet.exists?(account_id: account_id,
+                                          _blockchain: _blockchain,
+                                          primary_wallet: true)
+  end
+
+  def mark_another_wallet_as_primary
```
aquabu, line 45: I'm not sure about the intent of this function.

I would think that `mark_another_wallet_as_primary` would allow the caller to specify another wallet as the primary. This seems to set the earliest wallet as primary as an afterhook. Maybe rename the method to `mark_another_wallet_as_primary_after_primary_wallet_is_destroyed` - a bit long but maybe better.

app/services/wallet_creator.rb
```diff
@@ -1,25 +1,31 @@
...
 
   def initialize(account:)
     @account = account
   end
 
-  def call(wallet_params, tokens_to_provision: [])
-    @wallet = account.wallets.new(wallet_params)
-    @tokens_to_provision = sanitize_tokens_to_provision(tokens_to_provision)
+  def call(wallets_params)
+    wallets_params.map do |wallet_params|
```
aquabu, line 11: `wallets_params.map do |wallet_params|` uses the same name `wallet_params` as the array mapped over and the name of the element in the array.

maybe change to `wallets_params.map do |wallet_param|`

app/services/wallet_creator.rb
```diff
@@ -1,25 +1,31 @@
...
+      tokens_to_provision = wallet_params.delete(:tokens_to_provision)
```
aquabu, line 12: it's a little unclear what's happening here... is the wallet_params the array being mapped over? If so, deleting an element from the array being mapped over within the map is unclear. I would recommend keeping track of the errors as a second variable or exiting the loop without `wallet.save` if there's an error.

db/migrate/20210106184139_add_column_primary_wallet_to_wallets.rb
```diff
@@ -0,0 +1,27 @@
+class AddColumnPrimaryWalletToWallets < ActiveRecord::Migration[6.0]
+  def up
+    add_column :wallets, :primary_wallet, :boolean, default: false
+    remove_index :wallets, name: 'idx_blockchain_account_id'
+    add_index :wallets, [:account_id, :_blockchain]
+    add_index :wallets, [:account_id, :primary_wallet, :_blockchain], unique: true, where: "(primary_wallet IS TRUE)"
+
+    execute <<-SQL
```
aquabu, line 8: 1) Since there is only one wallet per blockchain network currently, couldn't this just be done by creating the `primary_wallet` with `default:true` in an earlier migration and then in a second commit changing the default to false?

2) Otherwise this should be done as a data migration and be run with `rake data:migrate` instead of `rake db:migrate`. 

I think it's probably better to do it as approach 1.

aquabu (changes requested):
I added a number of comments to resolve.

app/models/wallet.rb
Riddlerrr: I suggest name it `mark_first_wallet_as_primary` because it is exactly what we do in the method

app/controllers/api/v1/wallets_controller.rb
```diff
@@ -11,14 +11,42 @@ def index
...
-      render 'show.json', status: :created
-    else
-      @errors = @wallet.errors
+      if @wallets.all?(&:persisted?)
+        render 'index.json', status: :created
+      else
+        @errors = {}
 
-      render 'api/v1/error.json', status: :bad_request
+        @wallets.each_with_index.map do |wallet, i|
```
Riddlerrr, line 22:
```suggestion
         @wallets.map.with_index do |wallet, i|
```
It just reads a little bit better

app/services/wallet_creator.rb
```diff
@@ -1,25 +1,31 @@
...
 
-    build_wallet_provisions if should_be_provisioned?
+      wallet = account.wallets.new(wallet_params)
+      tokens_to_provision = sanitize_tokens_to_provision(wallet, tokens_to_provision)
 
-    wallet.save if wallet.errors.empty?
-    wallet
+      build_wallet_provisions(wallet, tokens_to_provision) if should_be_provisioned?(wallet, tokens_to_provision)
+
+      wallet.save if wallet.errors.empty?
```
Riddlerrr, line 20: I wonder can we just call `account.save` outside the loop to save all changes?

app/models/award.rb
matjacdev515: We already storing the final address inside the BlockchainTransactionAward model in the `destination` field, if I understand correctly

app/services/wallet_creator.rb
matjacdev515: Not sure that I understand. There two different namings `wallets_params` (plural) and `wallet_params` (singular).

`wallets_params` - array of hashes.
`wallet_params` - Hash, list of attributes per wallet.

`wallet_param` - looks more like a specific attribute, not a set of them

app/services/wallet_creator.rb
matjacdev515: `wallets_params` it is an array of params per wallet. I removing `tokens_to_provision`  from a wallet hash, because this attribute can't be passed to `account.wallets.new` and should be processed separately.

> I would recommend keeping track of the errors as a second variable or exiting the loop without `wallet.save` if there's an error.

I can't exit the loop because we need to provide error messages for all the wallets sent in the request. So we shouldn't fail on first. But this is a good idea not to try save particular wallet until we make sure that all the wallets can be saved and get rid of `ActiveRecord::Base.transaction` inside the controller.

db/migrate/20210106184139_add_column_primary_wallet_to_wallets.rb
matjacdev515: Good point

app/services/wallet_creator.rb
matjacdev515: Maybe we can play with naming little bit to make them look different e.g. `wallets_params` and `wallet_attrs`

app/services/wallet_creator.rb
matjacdev515: Will test this in different ways, since we are also building another relation inside a wallet `wallet.wallet_provisions.new` this can be a problem.

app/models/award.rb
aquabu: ok - let's rename the association to `recipient_wallet`

app/services/wallet_creator.rb
aquabu: ah, I didn't see the s on wallets_. Yes, more distinct naming would be better.

aquabu (approved):

Riddlerrr (approved):
Looks great 👍 
P.S. I didn't look through specs but I believe it is ok too.

PR #1811, opened 2021-01-11
AccountTokenRecord: Add wallet relation

Riddlerrr:
We decided to link AccountTokenRecord with wallet instead of Account because we will have the ability to have multiple wallets for blockchain. This change needs to be before we will continue with https://github.com/CoMakery/comakery-server/pull/1804

Also I've added `WalletTransferRule` as an alias for `AccountTokenRecord`. It's preparing for the task: https://www.pivotaltracker.com/story/show/176374185


aquabu:
Note: this should be merged after the create multiple wallet PR.

db/data_migrations/20210111085831_fill_wallet_for_account_token_record.rb
```diff
@@ -0,0 +1,12 @@
+class FillWalletForAccountTokenRecord < ActiveRecord::DataMigration
+  def up
+    AccountTokenRecord.where(wallet: nil).includes(:token, :reg_group).find_each do |atr|
```
aquabu, line 3: Could you write a test for this?

aquabu (changes requested):

db/data_migrations/20210111085831_fill_wallet_for_account_token_record.rb
Riddlerrr: 🤔 I think we don't have tools to test data migrations
@Alexey1100 Do you have any ideas about how we can add specs the data migration?

Riddlerrr: Added tests

Alexey1100 (approved):

aquabu (approved):

aquabu (approved):

PR #1814, opened 2021-01-12
Add full error details into OreIdService::Error

Riddlerrr:
It helps us to debug reasons for rejected/error transactions.
Was added when I debugged [this task](https://www.pivotaltracker.com/story/show/175423514)

### DIFF
--- a/app/models/ore_id_service.rb
+++ b/app/models/ore_id_service.rb
@@ -193,7 +193,7 @@ class OreIdService
       when 'userAlreadyExists'
         raise OreIdService::RemoteUserExistsError
       else
-        raise OreIdService::Error, "#{body['message']} (#{body['errorCode']} #{body['error']})"
+        raise OreIdService::Error, "#{body['message']} (#{body['errorCode']} #{body['error']})\nFull details: #{body}"
       end
     end
 


app/models/ore_id_service.rb
```diff
@@ -193,7 +193,7 @@ def raise_error(body)
       when 'userAlreadyExists'
         raise OreIdService::RemoteUserExistsError
       else
-        raise OreIdService::Error, "#{body['message']} (#{body['errorCode']} #{body['error']})"
+        raise OreIdService::Error, "#{body['message']} (#{body['errorCode']} #{body['error']})\nFull details: #{body}"
```
aquabu, line 196: Is it possible for this to leak any sensitive information?

aquabu (changes requested):

app/models/ore_id_service.rb
Riddlerrr: I don't think so. It will send to Sentry and not available for users and there is no sensitive info. An example:
↣ ruby  sergey@MacBook-Pro-Andronov local  2021-01-12 13-23-14

Alexey1100 (approved):

aquabu (approved):

PR #1816, opened 2021-01-12
Add validations to project images

Lerek:


aquabu (approved):

aquabu (approved):

app/models/project.rb
```diff
@@ -55,6 +55,9 @@ class Project < ApplicationRecord
   validate :valid_video_url, if: -> { video_url.present? }
   validate :token_changeable, if: -> { token_id_changed? && token_id_was.present? }
   validate :terms_should_be_readonly, if: -> { legal_project_owner_changed? || exclusive_contributions_changed? || confidentiality_changed? }
+  validates :image, dimension: { min: 1..1 }
```
Alexey1100, line 58: What happens if user tries to upload a video, which also might have dimensions..

PR #1820, opened 2021-01-14
Bug/fix Request fails when account id is nil instead of validations errors

matjacdev515:
Task https://www.pivotaltracker.com/story/show/176238550

### DIFF
--- a/spec/acceptance/api/v1/14_transfers_spec.rb
+++ b/spec/acceptance/api/v1/14_transfers_spec.rb
@@ -5,7 +5,7 @@ resource 'IV. Transfers' do
   include Rails.application.routes.url_helpers
 
   let!(:active_whitelabel_mission) { create(:mission, whitelabel: true, whitelabel_domain: 'example.org', whitelabel_api_public_key: build(:api_public_key), whitelabel_api_key: build(:api_key)) }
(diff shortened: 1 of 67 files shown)


spec/acceptance/api/v1/14_transfers_spec.rb
```diff
@@ -84,7 +87,7 @@
       parameter :total_amount, 'transfer total_amount (amount times quantity, same decimals as token)', required: true, type: :string
       parameter :account_id, 'transfer account id', required: true, type: :string
       parameter :transfer_type_id, 'custom transfer type id (default: earned)', required: false, type: :string
-      parameter :recipient_wallet_id, 'custom recipient wallet id', required: false, type: :string
+      parameter :recipient_wallet_id, 'custom recipient wallet id', required: false, type: :integer
```
Riddlerrr, line 90: All params must be strings. It is a requirement for calculating sign hash.

Riddlerrr (changes requested):
I think we need some more controller tests to check the behaviour with both cases: account present or not.

spec/acceptance/api/v1/14_transfers_spec.rb
matjacdev515: Got it

aquabu (approved):

PR #1821, opened 2021-01-15
Add primary wallet feature to web

Lerek:


Riddlerrr (changes requested):
Let's also use this new `MakePrimaryWallet` class into `Api::V1::WalletsController#update`:
https://github.com/CoMakery/comakery-server/blob/acceptance/app/controllers/api/v1/wallets_controller.rb#L27

app/controllers/primaries_controller.rb
```diff
@@ -0,0 +1,21 @@
+class PrimariesController < ApplicationController
```
Riddlerrr, line 1: 🤔 I think in this case better to simply add `set_primary` action to `WalletController`
It should improve readability. I can't guess what the controller does just by its name.

Lerek: @Riddlerrr maybe then it's better just to rename the controller to `WalletPrimariesController` to keep REST style, what do you think about it?

Riddlerrr: I thought about it too but my vote still to `set_primary` action. I don't think REST is so important in this case, it is not an API. And this controller will be always with just one action for sure.

@Alexey1100 What do you think?

Lerek: @Riddlerrr nvm, I do not insist :) I will move the logic to new action

Riddlerrr (approved):
👍  Looks good to me

Riddlerrr:
@aquabu It is ready to merge

aquabu (approved):

PR #1822, opened 2021-01-15
Feature/validate images content type and size

swimsoft:
- Restrict content_type and size for all attachments
- Add specs for all attachment validations
- Include active_storage_validations for testing the validators
- Generate data-migration to validate existing attachments
- Fix offenses
- Fix specs

### DIFF
--- a/app/models/mission.rb
+++ b/app/models/mission.rb
@@ -48,20 +48,20 @@ class Mission < ApplicationRecord
   validates :description, length: { maximum: 500 }
   validate :whitelabel_api_public_key_cannot_be_overwritten, if: -> { whitelabel_api_public_key_changed? && whitelabel_api_public_key_was.present? }
   validates :image, content_type: %w[image/png image/jpg image/jpeg],
-            size: { less_than: 10.megabytes , message: 'is not given between size' },
-            dimension: { min: 1..1 }
+                    size: { less_than: 10.megabytes, message: 'is not given between size' },
+                    dimension: { min: 1..1 }
   validates :logo, content_type: %w[image/png image/jpg image/jpeg],
-            size: { less_than: 10.megabytes , message: 'is not given between size' },
-            dimension: { min: 1..1 }
+                   size: { less_than: 10.megabytes, message: 'is not given between size' },
+                   dimension: { min: 1..1 }
   validates :whitelabel_logo, content_type: %w[image/png image/jpg image/jpeg],
-            size: { less_than: 10.megabytes , message: 'is not given between size' },
-            dimension: { min: 1..1 }
+                              size: { less_than: 10.megabytes, message: 'is not given between size' },
+                              dimension: { min: 1..1 }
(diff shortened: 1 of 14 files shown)


app/models/mission.rb
```diff
@@ -47,6 +47,21 @@ class Mission < ApplicationRecord
   validates :subtitle, length: { maximum: 140 }
   validates :description, length: { maximum: 500 }
   validate :whitelabel_api_public_key_cannot_be_overwritten, if: -> { whitelabel_api_public_key_changed? && whitelabel_api_public_key_was.present? }
+  validates :image, content_type: %w[image/png image/jpg image/jpeg],
```
Riddlerrr, line 50: Can we extract those settings into custom validator and reuse it everywhere? Because it's a lot of code duplication.

db/schema.rb
```diff
@@ -60,12 +60,12 @@
     t.string "image_content_size"
     t.string "image_content_type"
     t.string "nickname"
-    t.string "country"
-    t.date "date_of_birth"
     t.string "public_address"
     t.string "nonce"
     t.string "network_id"
     t.boolean "system_email", default: false
+    t.string "country"
```
Riddlerrr, line 67: Could you revert the schema changes? There are no schema changes in this PR.
Also, looks like you loaded DB when the schema was wrong so I suggest re-create your dev DB with the fresh schema.

Riddlerrr (changes requested):

db/data_migrations/20210114220605_validate_attachments.rb
```diff
@@ -0,0 +1,23 @@
...
+    mission: %w[logo image whitelabel_logo whitelabel_logo_dark whitelabel_favicon],
+    project: %w[image square_image panoramic_image],
+    token: %w[log_image]
+  }.freeze
+
+  def up
+    MODEL_AND_FIELDS.each do |model, fields|
+      model.constantize.find_each do |m|
+        fields.each do |field|
+          m.send(field).purge if m.send(field).attached? && is_valid?(field)
```
Alexey1100, line 14: Should it be?

```suggestion
          m.send(field).purge if m.send(field).attached? && invalid?(field)
```

db/data_migrations/20210114220605_validate_attachments.rb
```diff
@@ -0,0 +1,23 @@
...
+        end
+      end
+    end
+  end
+
+  def valid?(field)
```
Alexey1100, line 20:
```suggestion
  def invalid?(field)
```
👀

app/models/mission.rb
Alexey1100: Yep, and we could move all of the specs into a shared example which accepts list of column names and iterates over it (ref `belongs_to_blockchain_spec.rb:30`)

Alexey1100 (approved):
👍

Riddlerrr (approved):
Perfect! 👍

PR #1823, opened 2021-01-18
Add uniqueness validation to wallet address

Lerek:


Riddlerrr (changes requested):
I think this task require adding unique DB index too and check and remove already exists duplicates first.
It can be implemented in one migration.

Lerek:
@Riddlerrr well, there already could some wallet with the same addres-blockchain exist... Should I create migration to destroy duplicates first?

Riddlerrr:
Nope, I think will be fine to do it in the same migration. You can just remove duplicates before adding the index in the same migration.
Data migrations will also not fit in this case because it will run _after_ regular migrations but we must ensure what we don't have any duplicates _before_ creating the index.

Lerek:
@Riddlerrr done, could you please take another look?

db/migrate/20210118172946_add_address_blockchain_account_index_to_wallets.rb
```diff
@@ -0,0 +1,11 @@
+class AddAddressBlockchainAccountIndexToWallets < ActiveRecord::Migration[6.0]
+  def up
+    ids_to_keep = Wallet.select('DISTINCT ON (account_id, address, _blockchain) *').map(&:id)
+    Wallet.where.not(id: ids_to_keep).destroy_all
+    add_index :wallets, [:account_id, :address, :_blockchain], unique: true
+  end
+
+  def down
+    emove_index :wallets, [:account_id, :address, :_blockchain]
```
Alexey1100, line 9:
```suggestion
    remove_index :wallets, [:account_id, :address, :_blockchain]
```

Riddlerrr (approved):
👍

PR #1824, opened 2021-01-18
Add pagination to tokens page

Lerek:


Riddlerrr:
The code looks good to me 👍

Would be nice to have a screenshot in PR description if it related to frontend changes to show the new version.
It also will help to test the changes.

Lerek:
@Riddlerrr thanks, and sure, here it is:
image

app/controllers/tokens_controller.rb
```diff
@@ -81,6 +81,25 @@ def fetch_contract_details
 
   private
 
+    def set_tokens
```
matjacdev515, line 84:
```ruby
def tokens
    @tokens ||=
      policy_scope(Token)
      .with_attached_logo_image
      .page(params[:page])
      .per(TOKENS_PER_PAGE)
  end
```

And use `tokens` method where it is needed, without calling `set_tokens` separately. What do you think?

Lerek:
@matjacdev515 done, could you please take another look? :)

aquabu (approved):

PR #1833, opened 2021-01-21
API: Introduce '/api/v1/wallet_recovery/public_wrapping_key' endpoint

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/176549183

### DIFF
--- a/app/controllers/api/v1/wallet_recovery_controller.rb
+++ b/app/controllers/api/v1/wallet_recovery_controller.rb
@@ -2,8 +2,8 @@ class Api::V1::WalletRecoveryController < Api::V1::ApiController
   # TODO: Which authorization use here?
 
   def public_wrapping_key
-    public_key = Eth::Key.new(priv: ENV['WALLET_RECOVERY_WRAPPING_KEY']).public_key.key
-
+    private_key = ENV['WALLET_RECOVERY_WRAPPING_KEY'] || 'default_key'
+    public_key = Eth::Key.new(priv: private_key).public_key.key
     render json: { public_wrapping_key: public_key }
   rescue MoneyTree::Key::KeyFormatNotFound
     @errors = { invalid_env_variable: 'WALLET_RECOVERY_WRAPPING_KEY variable was not configured or use wrong format' }
--- a/app/controllers/api/v1/wallet_recovery_controller.rb
+++ b/app/controllers/api/v1/wallet_recovery_controller.rb
@@ -1,8 +1,11 @@
 class Api::V1::WalletRecoveryController < Api::V1::ApiController
-  # TODO: Which authorization use here?
+  include Api::V1::Concerns::AuthorizableByMissionKey
+  include Api::V1::Concerns::RequiresAnAuthorization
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/wallet_recovery_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Api::V1::WalletRecoveryController < Api::V1::ApiController
+  # TODO: Which authorization use here?
```
Alexey1100, line 2: I think regular set of authorisations:

- requires header
- requires auth
- auth by signature proof

app/controllers/api/v1/wallet_recovery_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  def public_wrapping_key
+    private_key = ENV['WALLET_RECOVERY_WRAPPING_KEY'] || 'default_key'
+    public_key = Eth::Key.new(priv: private_key).public_key.key
```
Alexey1100, line 6: I think we need to use `ed25519` here (ref `Comakery::APISignature`)

spec/controllers/api/v1/wallet_recovery_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.describe Api::V1::WalletRecoveryController, type: :controller do
+  describe 'GET /public_wrapping_key' do
+    render_views
+
+    context 'with correct private key' do
+      # To check you can use https://gobittest.appspot.com/Address
+      it 'renders a successful response' do
+        ENV['WALLET_RECOVERY_WRAPPING_KEY'] = '18E14A7B6A307F426A94F8114701E7C8E774E7F9A47E2C2035DB29A206321725'
```
matjacdev515, line 10: The tricky thing about this way of setting the value of ENV variable, that the value will not be rolled back on test finish. I saw the same things in different tests, maybe we need to consider using Rspec mocks for that instead

app/controllers/api/v1/wallet_recovery_controller.rb
Riddlerrr: In the task description mentioned `secp256k1`
@aquabu Could you confirm which crypto method we must use in this case?

spec/controllers/api/v1/wallet_recovery_controller_spec.rb
Riddlerrr: I agree it's bad to rewrite ANV vars in tests:
Changed it to:
```ruby
allow(ENV).to receive(:[]).with('WALLET_RECOVERY_WRAPPING_KEY').and_return('18E14A7B6A307F426A94F8114701E7C8E774E7F9A47E2C2035DB29A206321725')
```

app/controllers/api/v1/wallet_recovery_controller.rb
Riddlerrr: Ok, confirmed using `secp256k1` through Ethereum lib

aquabu (approved):

PR #1835, opened 2021-01-21
Remove missions created by automation

swimsoft:
- Remove missions created by automation
- Fix offenses

### DIFF
--- a/db/data_migrations/20210121113148_remove_missions_created_by_automation.rb
+++ b/db/data_migrations/20210121113148_remove_missions_created_by_automation.rb
@@ -2,7 +2,7 @@ class RemoveMissionsCreatedByAutomation < ActiveRecord::DataMigration
   def up
     if Rails.env.production?
       name = Mission.arel_table[:name]
-      Mission.where(name.matches("%#{'Automation'}%")).destroy_all
+      Mission.where(name.matches('%Automation%')).destroy_all
     end
   end
 end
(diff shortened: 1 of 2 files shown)


db/data_migrations/20210121113148_remove_missions_created_by_automation.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveMissionsCreatedByAutomation < ActiveRecord::DataMigration
+  def up
+    if Rails.env.production?
+      name = Mission.arel_table[:name]
```
Alexey1100, line 4: Let's add a spec here :)

E.g: `spec/data_migrations/fill_wallet_for_account_token_record_spec.rb`

Alexey1100 (dismissed):
Specs

db/data_migrations/20210121113148_remove_missions_created_by_automation.rb
```diff
@@ -0,0 +1,8 @@
...
+      Mission.where(name.matches('%Automation%')).destroy_all
```
matjacdev515, line 5: Think `Mission.where("name LIKE '%Automation%'").destroy_all` will be more developer-friendly)

Riddlerrr (approved):

db/data_migrations/20210121113148_remove_missions_created_by_automation.rb
aquabu: @swimsoft I'm going to merge without resolving this comment from @matjacdev515 to keep the code moving. But I think @matjacdev515. Can you add the change to a future commit?

PR #1838, opened 2021-01-25
Add PM2 process manager and boilerplate for Hotwallet process

Lerek:


hotwallet/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "hotwallet",
+  "version": "1.0.0",
+  "description": "Hotwallet process",
+  "main": "app.js",
+  "license": "MIT",
```
aquabu, line 6: License should be `Noncommercial Nonviolent Public License 1.1` to match the comakery-server repository. May change this to MIT for the hot wallet later but we would have to clarify which license applies to which code more carefully.

aquabu (approved):

PR #1840, opened 2021-01-25
Add wallet recovery public api key to missions

Lerek:


Riddlerrr:
I think the `wallet_recovery_api_public_key` should be available to set using API only and not available to show/edit using UI.
But let's check with @aquabu

aquabu:
This is the public key used for signing API requests similar to the Whitelabel API key.

You should be able to add it from the missions page. But it should not be editable after it is posted. Currently you would need to contact a site administrator to change it after it has been posted.

I don't think there is an API endpoint for this and there probably shouldn't be for now.

app/javascript/components/MissionForm.jsx
```diff
@@ -365,6 +397,14 @@ export default class MissionForm extends React.Component {
             eventHandler={this.handleChangeFormData}
           />
 
+          <InputFieldWhiteDark
+          name='walletRecoveryApiPublicKey'
+          title='Wallet recovery API Public Key'
+          symbolLimit={68}
+          value={walletRecoveryApiPublicKey}
```
Alexey1100, line 404:
```suggestion
          value={walletRecoveryApiPublicKey}
          readOnly={this.props.mission.walletRecoveryApiPublicKey}
```

Alexey1100 (changes requested):
Should be read-only once set

app/javascript/components/MissionForm.jsx
aquabu: Yes and the read only property should be on the model level so the form can't be bypassed.

Lerek:
@aquabu , @Alexey1100 done, could you take a look again, please?

Riddlerrr (changes requested):
Let'a add `wallet_recovery_api_public_key` as a readonly attribute for the model. Like we do for example here:
https://github.com/CoMakery/comakery-server/blob/b7aa5316cd70b03051592d841543fee26defcf9a/app/models/reg_group.rb#L15

Lerek:
> Let'a add `wallet_recovery_api_public_key` as a readonly attribute for the model. Like we do for example here:
> https://github.com/CoMakery/comakery-server/blob/b7aa5316cd70b03051592d841543fee26defcf9a/app/models/reg_group.rb#L15

@Riddlerrr done

Riddlerrr (approved):

aquabu:
@Lerek I think this issues are fixed but there's a merge conflict.

Lerek:
@aquabu done)

Alexey1100 (approved):

PR #1842, opened 2021-01-26
Add email field to form after signuping via metamask

Lerek:
image


Riddlerrr (approved):

aquabu (changes requested):
1. Needs a spec. 
2. This should only be shown after signup with MetaMask because. The other signup processes provide the email on the previous signup page.

Lerek:
> 1. Needs a spec.
> 2. This should only be shown after signup with MetaMask because. The other signup processes provide the email on the previous signup page.

done

aquabu (approved):

PR #1847, opened 2021-01-27
Update account token record api endpoints

Lerek:


Riddlerrr:
@Lerek When I looked through your code in the PR I've noticed what our initial task was not well thought out.
You did everything right. We just forgot that that now `account_token_records` can have multiple accounts. It complicates things and we decided to leave only `wallet_id` for `show` and `delete` actions.

Updated:
INDEX /api/v1/tokens/: token_id/account_token_records
INDEX /api/v1/tokens/: token_id/account_token_records?account_id=:account_id
CREATE /api/v1/tokens/: token_id/account_token_records
GET /api/v1/tokens/:token_id/account_token_records/?wallet_id=:wallet_id
DELETE /api/v1/tokens/:token_id/account_token_records/?wallet_id=:wallet_id

Riddlerrr:
@Alexey1100 Please double check the new URLs above ☝️

Lerek:
> @Alexey1100 Please double check the new URLs above ☝️

@Riddlerrr okey, I've adjusted to the updated toutes list)

app/controllers/api/v1/account_token_records_controller.rb
```diff
@@ -1,55 +1,67 @@
...
+
+    def token
+      @token ||= Token.find(params[:token_id])
     end
 
-    def account_token_record
-      @account_token_record ||= project.account_token_records.find(params[:id])
+    def find_account_from_body
+      account_id = params.dig(:body, :data, :account_token_record, :account_id)
+      Account.find(account_id)
```
Alexey1100, line 64: We are using managed_account_id instead of real one in whitelabel API requests.
So, looks like we need to update this controller as well.

spec/acceptance/api/v1/16_account_token_records_spec.rb
```diff
@@ -6,73 +6,60 @@
...
-      response_field :id, 'account token record id', type: :integer
-      response_field :account_id, 'account id', type: :integer
-      response_field :token_id, 'token id', type: :integer
-      response_field :reg_group_id, 'reg group id', type: :integer
-      response_field :lockup_until, 'lockup until', type: :integer
-      response_field :max_balance, 'max balance', type: :integer
-      response_field :account_frozen, 'account frozen', type: :bool
-      response_field :status, 'account token record status (created synced)', type: :string
-      response_field :createdAt, 'creation timestamp', type: :string
-      response_field :updatedAt, 'update timestamp', type: :string
```
Alexey1100, line 56: We need to keep response fields reference in the documentation

Alexey1100 (changes requested):
@Lerek I think we need a couple more updates here, as mentioned below.

Also, I noticed that `account-form_controller.js:13` is going to be broken, because it was using AccountTokenRecord API endpoints with project policy auth.

I think the best solution atm would be moving away from using that API endpoint with Metamask and creating a separate controller `project/:id/dashboard/account_token_records` with only one `create` action and `project#edit?` policy, and update the stimulus controller to use that action.

spec/acceptance/api/v1/16_account_token_records_spec.rb
Lerek: @Alexey1100 but there is no such endpoint anymore, that's why it's been removed

app/controllers/api/v1/account_token_records_controller.rb
Lerek: @Alexey1100 sorry, seems i'm not following, this portion of code was just a bit refactored and reflects exactly the same logic as it was before. Are you sure here should be changed something, and if so, what exactly?

Lerek:
> @Lerek I think we need a couple more updates here, as mentioned below.
> 
> Also, I noticed that `account-form_controller.js:13` is going to be broken, because it was using AccountTokenRecord API endpoints with project policy auth.
> 
> I think the best solution atm would be moving away from using that API endpoint with Metamask and adding a `create` action into `Dashboard::AccountsController` with `project#edit?` policy, and update the stimulus controller to use that action.

@Alexey1100 maybe atm it's easier just to update the path passed to the stimulus controller, like I did it here: [changed endpoint](https://github.com/CoMakery/comakery-server/pull/1847/files#diff-2558c3b9a7c17301bf2e76e2b3a8df944d8528bd72b5bbb84c6e12e8880333e5R6)?

Alexey1100:
@Lerek The issue is that we also removed project policy auth from the api controller, which is not associated with project anymore and requires a full signature auth.

spec/acceptance/api/v1/16_account_token_records_spec.rb
Alexey1100: Yes, we return this fields in INDEX response now.

app/controllers/api/v1/account_token_records_controller.rb
Alexey1100: @Lerek Yep, it's a bit over complicated for security reasons, but we are using `Account#managed_account_id` instead of real `Account#id` in whitelabel API, you can look at other account-related endpoints for the reference.

Since this endpoint is updated and no longer used by stimulus controller, we should switch to `managed_account_id ` here as well.
We don't expose the real account id anywhere else, so it was a bit tricky to use this endpoint from API user perspective.

Lerek:
@Alexey1100 done

Alexey1100 (approved):
👍

PR #1848, opened 2021-01-27
Feature/api endpoint create hot wallet address

swimsoft:
- Create endpoint to register hot wallet address with a project
- Fix offenses
- Generate new documentation
- Change authorization. update specs
- Fix specs
- Fix conflicts


app/controllers/api/v1/hot_wallet_addresses_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    else
+      @errors = @build_hot_wallet.errors
+      render 'api/v1/error.json', status: :bad_request
+    end
+  end
+
+  private
+
+    def build_hot_wallet
+      @build_hot_wallet ||= project.create_hot_wallet(wallet_params)
```
aquabu, line 26: The method name build seems like it implies building the object in memory but not persisting it to the database? It seems like the `build_hot_wallet` and `project.create_hot_wallet` method names should match. They also should be named `build_not_wallet` if only in memory or `create_hot_wallet` when persisted to the database.

spec/controllers/api/v1/hot_wallet_addresses_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        params[:project_id] = project.id
+        params[:format] = :json
+
+        post :create, params: params
+        expect(response).to have_http_status(:created)
+        expect(project.reload.hot_wallet.present?).to be true
+      end
+    end
+
+    context 'when project has assign a hot wallet' do
```
aquabu, line 43: Minor typo:
'when project already has an assigned hot wallet'

aquabu (changes requested):

app/controllers/api/v1/hot_wallet_addresses_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  include Api::V1::Concerns::AuthorizableByMissionKey
+  include Api::V1::Concerns::RequiresAnAuthorization
+  include Api::V1::Concerns::RequiresSignature
+  include Api::V1::Concerns::RequiresWhitelabelMission
+
+  before_action :verify_hot_wallet, only: :create
+
+  def create
+    build_hot_wallet
+    @build_hot_wallet.source = Wallet.sources[:hot_wallet]
```
Riddlerrr, line 11: minor comment:
`@build_hot_wallet.source = :hot_wallet`?

swimsoft: 👍

swimsoft: 👍

app/controllers/api/v1/hot_wallet_addresses_controller.rb
aquabu: @swimsoft I'm going to merge this but could you create a follow up commit that addresses this naming issue? Thanks.

aquabu (approved):

PR #1850, opened 2021-01-28
Cover by test the case when nonce was already used

Riddlerrr:
The quite important line was not covered:
Code coverage for Comakery-server 2021-01-28 14-03-16

This PR fix that

image


Alexey1100 (approved):
👍

aquabu (approved):

PR #1851, opened 2021-01-29
API: raise an error if can not write into cache

Riddlerrr:
Task: https://www.pivotaltracker.com/story/show/176699487

> 1. Raise an error if Rails.cache.write returns false
> 2. Trim provided nonce to 36 characters (UUID length), before processing

### DIFF
--- a/app/lib/cache_extension.rb
+++ b/app/lib/cache_extension.rb
@@ -1,0 +1,14 @@
+module CacheExtension
+  class WriteFailed < StandardError; end
+
+  module_function
+
+  def cache_write!(key, value, options = {})
+    write_result = Rails.cache.write(key, value, **options) # will return nil even it fails, details: https://guides.rubyonrails.org/v6.0/caching_with_rails.html#activesupport-cache-rediscachestore
+
+    # failed to write
+    raise WriteFailed, "Can't write into cache: #{key} = #{value}" unless write_result
+
+    true
+  end
+end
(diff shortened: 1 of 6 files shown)


app/lib/cache_extension.rb
```diff
@@ -0,0 +1,14 @@
+module CacheExtension
+  class WriteFailed < StandardError; end
+
+  module_function
+
+  def cache_write!(key, value, options = {})
+    write_result = Rails.cache.write(key, value, **options) # will return nil even it fails, details: https://guides.rubyonrails.org/v6.0/caching_with_rails.html#activesupport-cache-rediscachestore
+
+    # failed to write
+    raise WriteFailed, "Can't write into cache: #{key} = #{value}" unless write_result
```
Alexey1100, line 10: `Rails.cache.as_json` outputs cache max size and current size.
I think we can include that into exception message.

Screenshot 2021-01-29 at 22 20 07

Alexey1100 (approved):

app/lib/comakery/api_signature.rb
```diff
@@ -164,7 +166,7 @@ def verify_http_method
       end
 
       def verify_nonce
-        raise Comakery::APISignatureError, 'Invalid nonce' unless @is_nonce_unique.call(@request.fetch('body', {}).fetch('nonce', ''))
+        raise Comakery::APISignatureError, 'Invalid nonce' unless @is_nonce_unique.call(normalize_nonce(@request.fetch('body', {}).fetch('nonce', '')))
```
aquabu, line 169: We should probably return an error if the nonce size is > MAX_NONCE_SIZE. Otherwise the behavior could be odd could be odd for API users. 

For example if I used a long prefix for the nonce it would truncate the random part of the nonce. Then I would get intermittent nonce failures even though I was generating sufficiently random nonces on the client side.

aquabu (approved):

PR #1853, opened 2021-01-29
Feature/hot wallet address is visible on transfers page

swimsoft:
Task: https://www.pivotaltracker.com/story/show/176622298
Screenshot_1


aquabu (approved):

app/controllers/api/v1/hot_wallet_addresses_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  include Api::V1::Concerns::RequiresAnAuthorization
+  include Api::V1::Concerns::RequiresSignature
+  include Api::V1::Concerns::RequiresWhitelabelMission
+
+  before_action :verify_hot_wallet, only: :create
+
+  def create
+    build_hot_wallet
+    @build_hot_wallet.source = Wallet.sources[:hot_wallet]
+    @build_hot_wallet.account = project.account
```
Alexey1100, line 12: I don't think this is the implementation we needed.
In this case the wallet is associated with the project owner and will be visible on wallets page for that account.

What we need to is to make wallet belong to project instead of account if source is `hot_wallet`

@swimsoft @aquabu @Riddlerrr what do you think?

Alexey1100: Sorry, probably we needed to describe it more verbose.

aquabu: yes, true. The hot wallet will be unusual because it does not have an owner.

Alexey1100: So, what we need to do now is:

1) Add project_id key to wallets table
2) Make db migration to revert the changes on projects table
3) Wallet belong to project if source is hot_wallet
4) Project has_one hot wallet

aquabu: I think we won't need the db migration (2) because we won't test it and create data to migrate. Or if there is one entry on demo we can manage it manually as a one off event.

Alexey1100: @aquabu we need to remove created wallet_id foreign key from projects table

PR #1854, opened 2021-01-29
Integrate Turbo & Dynamically update transfers table

matjacdev515:


app/javascript/controllers/account-form_controller.js
```diff
@@ -64,6 +64,13 @@ export default class extends ComakerySecurityTokenController {
     })
   }
 
+  _markButtonAsPending() {
+    this.buttonTarget.parentElement.classList.add('in-progress--metamask__paid')
+    this.buttonTarget.getElementsByTagName('span')[0].textContent = 'processing'
+
+    this.formTarget.requestSubmit()
```
Alexey1100, line 71: 👍

So, the logic is triggered and turbo frame got updated after the tx submitted to blockchain, right?

app/views/dashboard/accounts/_account.html.erb
```diff
@@ -2,171 +2,172 @@
 <% synced_record = account.account_token_records.where(token: @project.token, status: :synced).last %>
 <% last_record = account.account_token_records.where(token: @project.token).last %>
 <% last_record_is_not_synced = synced_record && last_record && synced_record.id != last_record.id %>
-<% form_url = last_record ? project_dashboard_account_path(@project, last_record) : root_path %>
+<% form_url = last_record ? project_dashboard_account_path(@project, account) : root_path %>
```
Alexey1100, line 5: There was some controller refactoring regarding the way account token records are created recently.
Pls merge acceptance and update the urls.

app/views/dashboard/accounts/_account.html.erb
```diff
@@ -2,171 +2,172 @@
...
 <% form_data = last_record ? { controller: "account-form", target: "account-form.form", 'account-form-address' => account.address_for_blockchain(@project.token._blockchain), 'account-form-account-id' => account.id, 'account-form-account-token-records-path' => api_v1_project_account_token_records_path(project_id: @project.id), 'account-form-transactions-path' => api_v1_project_blockchain_transactions_path(project_id: @project.id)}.merge(@project.token.decorate.eth_data('account-form')) : nil %>
```
Alexey1100, line 6: `api_v1_project_account_token_records_path ` is no longer available

app/views/tokens/new.html.erb
```diff
@@ -0,0 +1 @@
+<%= react_component('TokenForm', @props, prerender: false) %>
```
Alexey1100, line 1: Is there some issues with pre rendering this component?

Alexey1100 (changes requested):
@matjacdev515 Great work!
Pls merge acceptance and check that the account token record creation logic with Metamask is still works properly, since there were some critical updates recently.

Tests for that are running externally by @miemar and not currently included in the repo (since they require complicated setup with browser extension and local blockchain node), so it might be broken even if CI passes.

app/javascript/controllers/account-form_controller.js
matjacdev515: Yes, Turbo triggers only on a form submit or clicking a link

app/views/dashboard/accounts/_account.html.erb
matjacdev515: It is not API)

app/views/tokens/new.html.erb
matjacdev515: Yes, `render component:` not working with Turbo

app/views/dashboard/accounts/_account.html.erb
matjacdev515: @Alexey1100 it was changed to `project_dashboard_accounts_path`. But since it is not a API controller, this form rejected by CSRF because we still submitting it in `account-form_controller.js` by JS

```
Started POST "/projects/6/dashboard/accounts?token_id=7" for 172.22.0.1 at 2021-02-03 15:42:10 +0000
Cannot render console from 172.22.0.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
Processing by Dashboard::AccountsController#create as */*
  Parameters: {"body"=>{"data"=>{"account_token_record"=>{"max_balance"=>"0", "lockup_until"=>"1609545600", "reg_group_id"=>"0", "account_frozen"=>"true", "account_id"=>"1"}}}, "token_id"=>"7", "project_id"=>"6", "account"=>{}}
Can't verify CSRF token authenticity.
Completed 422 Unprocessable Entity in 2ms (ActiveRecord: 0.0ms | Allocations: 673)
```

Not sure in which way it should be fixed because it is not related to my changes

Alexey1100: I think we have two options:
1) Disable CSRF check for the endpoint
2) Generate CSRF token in the stimulus controller

app/views/tokens/new.html.erb
Alexey1100: Seems like on other pages it works fine

matjacdev515: Turbo sends specific Content-Type, something like `turbo-stream/html` which `render component:` doesn't know how to render and returns empty response.

matjacdev515: Sorry, I misunderstood.  `prerender: false` there because initially it was set `prerender: false` in the controller. So I didn't change anything. I'm not tested if prerender will work because this PR already has lost of changes and it's not really related to its subject.

Alexey1100 (approved):

spec/features/pages/featured_spec.rb
```diff
@@ -18,8 +18,11 @@
     login account
 
     visit root_path
-    expect(page).not_to have_content 'default project 8344'
-    expect(page).to have_content 'featured project 9934'
+    sleep 10
```
aquabu, line 21: @matjacdev515 I think this sleep and the save page needs to be taken out of the spec.

aquabu (approved):

PR #1859, opened 2021-02-02
Fix issue on register hot wallet address api

swimsoft:
Task: https://www.pivotaltracker.com/story/show/176550376

1. Add project_id key to wallets table
2. Revert the changes on projects table
3. Validate project_id to be assigned only if source is hot_wallet


aquabu:
@swimsoft there's a merge conflict to resolve on this.

app/models/project.rb
```diff
@@ -15,7 +15,7 @@ class Project < ApplicationRecord
   has_one_attached :panoramic_image
 
   belongs_to :account, touch: true
```
Alexey1100, line 17: Let's add `optional: true` here as well so it will be clear, and afaik default behaviour since rails 5 requires that.
We are probably using 4.x config, that's why it currently works.

app/models/wallet.rb
```diff
@@ -3,12 +3,12 @@ class Wallet < ApplicationRecord
   include BelongsToOreId
 
   belongs_to :account
```
Alexey1100, line 5: Let's add optional: true here so it will be clear, and afaik default behaviour since rails 5 requires that.
We are probably using 4.x config, that's why it currently works.

app/models/wallet.rb
```diff
@@ -30,6 +32,10 @@ def available_blockchains
     available_blockchains.map(&:key)
   end
 
+  def validate_project_id
```
Alexey1100, line 35: And we need to extend the validation here, to check that either account or project ids are present, depending on source.

Alexey1100 (changes requested):
Wallet validation logic updates

Alexey1100 (approved):

aquabu (approved):

PR #1867, opened 2021-02-05
Feature/transfer page chart balances display correctly

swimsoft:
Task: https://www.pivotaltracker.com/story/show/175048818

Fix matching colors
Add migration to fix source for awards
Rename Transfer Type to Category
Fix chart to support negative values

Screen Shot 2021-02-05 at 9 48 47 AM

### DIFF
--- a/app/controllers/dashboard/transfers_controller.rb
+++ b/app/controllers/dashboard/transfers_controller.rb
@@ -79,7 +79,7 @@ class Dashboard::TransfersController < ApplicationController
     end
 
     def transfer_type_name
-      @transfer_type_name ||= TransferType.find_by(id: params.fetch(:q, {}).fetch(:transfer_type_id_eq, nil))&.name
+      @transfer_type_name ||= TransferType.find_by(id: params.dig(:q, :transfer_type_id_eq))&.name
     end
 
     def transfer_params
--- a/app/controllers/dashboard/transfers_controller.rb
+++ b/app/controllers/dashboard/transfers_controller.rb
@@ -27,7 +27,6 @@ class Dashboard::TransfersController < ApplicationController
     @transfer.name = name.titlecase
     @transfer.account_id = params[:award][:account_id]
     @transfer.issuer = current_account
-    @transfer.source = Award.sources[name.to_sym] || 0
     @transfer.status = :accepted
 
     if @transfer.save
--- a/app/controllers/dashboard/transfers_controller.rb
+++ b/app/controllers/dashboard/transfers_controller.rb
@@ -24,7 +24,7 @@ class Dashboard::TransfersController < ApplicationController
     @transfer = @award_type.awards.new(transfer_params)
     name = @transfer.transfer_type.name
 
-    #TODO: I am using source to filter transfers but before source was set by default earned for all transfer types, is any logic behind?
+    # TODO: I am using source to filter transfers but before source was set by default earned for all transfer types, is any logic behind?
     @transfer.name = name.titlecase
     @transfer.account_id = params[:award][:account_id]
     @transfer.issuer = current_account
@@ -81,7 +81,7 @@ class Dashboard::TransfersController < ApplicationController
     end
 
     def transfer_type_name
-      @transfer_type ||= TransferType.find_by(id: params.fetch(:q, {}).fetch(:transfer_type_id_eq, nil))&.name
+      @transfer_type_name ||= TransferType.find_by(id: params.fetch(:q, {}).fetch(:transfer_type_id_eq, nil))&.name
     end
 
     def transfer_params
--- a/app/controllers/dashboard/transfers_controller.rb
+++ b/app/controllers/dashboard/transfers_controller.rb
@@ -24,7 +24,6 @@ class Dashboard::TransfersController < ApplicationController
     @transfer = @award_type.awards.new(transfer_params)
     name = @transfer.transfer_type.name
 
-    # TODO: I am using source to filter transfers but before source was set by default earned for all transfer types, is any logic behind?
     @transfer.name = name.titlecase
     @transfer.account_id = params[:award][:account_id]
     @transfer.issuer = current_account
--- a/app/controllers/dashboard/transfers_controller.rb
+++ b/app/controllers/dashboard/transfers_controller.rb
@@ -28,7 +28,7 @@ class Dashboard::TransfersController < ApplicationController
     @transfer.name = name.titlecase
     @transfer.account_id = params[:award][:account_id]
     @transfer.issuer = current_account
-    @transfer.source = name.to_sym
+    @transfer.source = Award.sources[name.to_sym] || 0
     @transfer.status = :accepted
 
     if @transfer.save
--- a/app/controllers/dashboard/transfers_controller.rb
+++ b/app/controllers/dashboard/transfers_controller.rb
@@ -22,9 +22,13 @@ class Dashboard::TransfersController < ApplicationController
     authorize @project, :create_transfer?
 
     @transfer = @award_type.awards.new(transfer_params)
-    @transfer.name = @transfer.transfer_type.name.titlecase
+    name = @transfer.transfer_type.name
+
+    #TODO: I am using source to filter transfers but before source was set by default earned for all transfer types, is any logic behind?
+    @transfer.name = name.titlecase
     @transfer.account_id = params[:award][:account_id]
     @transfer.issuer = current_account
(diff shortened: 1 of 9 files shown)


app/controllers/dashboard/transfers_controller.rb
```diff
@@ -22,9 +22,13 @@ def create
     authorize @project, :create_transfer?
 
     @transfer = @award_type.awards.new(transfer_params)
-    @transfer.name = @transfer.transfer_type.name.titlecase
+    name = @transfer.transfer_type.name
+
+    # TODO: I am using source to filter transfers but before source was set by default earned for all transfer types, is any logic behind?
```
aquabu, line 27: Tasks use the same model as transfers. Tasks have a default transfer type of earned. Please remove the comment in a future PR if the question is resolved.

app/decorators/project_decorator.rb
```diff
@@ -162,18 +162,18 @@ def transfers_chart_types
...
     project.transfer_types.map.with_index { |t, i| [t, Comakery::ChartColors.lookup(i)] }.to_h
   end
 
   # rubocop:todo Metrics/CyclomaticComplexity
-  def transfers_stacked_chart(transfers, limit, grouping, date_modifier, empty)
+  def transfers_stacked_chart(transfers, limit, grouping, date_modifier, empty, negative)
     chart = transfers.includes([:transfer_type]).where('awards.created_at > ?', limit).group_by { |r| r.created_at.send(grouping) }.map do |timeframe, set|
       transfers_chart_types.merge(
-        set.group_by(&:transfer_type).map { |k, v| [k.name, v.sum(&:total_amount)] }.to_h.merge(
+        set.group_by(&:transfer_type).map { |k, v| [k.name, negative ? -v.sum(&:total_amount) : v.sum(&:total_amount)] }.to_h.merge(
```
aquabu, line 176: This should probably be refactored into a separate function with it's own unit test to reduce cyclomatic complexity and increase reliability.

app/decorators/project_decorator.rb
```diff
@@ -184,43 +184,47 @@ def transfers_stacked_chart(transfers, limit, grouping, date_modifier, empty)
...
 
-  def transfers_stacked_chart_year(transfers)
+  def transfers_stacked_chart_year(transfers, negative: false)
     transfers_stacked_chart(
       transfers,
       10.years.ago,
       :beginning_of_year,
       '%Y',
-      (10.years.ago.year..Time.current.year).map { |k| transfers_chart_types.merge(timeframe: k.to_s, i: DateTime.strptime(k.to_s, '%Y').to_i) }
+      (10.years.ago.year..Time.current.year).map { |k| transfers_chart_types.merge(timeframe: k.to_s, i: DateTime.strptime(k.to_s, '%Y').to_i) },
```
aquabu, line 193: Seems like this should be refactored into it's own separate function with it's own unit test.

db/data_migrations/20210205131554_fix_source_for_awards.rb
```diff
@@ -0,0 +1,7 @@
+class FixSourceForAwards < ActiveRecord::DataMigration
+  def up
+    Award.find_each do |award|
+      award.update(:source, Award.sources[award.transfer_type.name.to_sym] || 0)
```
aquabu, line 4: What is the function of this? It also should have a test.

app/lib/comakery/chart_colors.rb
```diff
@@ -1,11 +1,12 @@
 class Comakery::ChartColors
+  # TODO: Check colors if are ordered correctly
```
aquabu, line 2: Please check this and remove the TODO.

aquabu (changes requested):

db/data_migrations/20210205131554_fix_source_for_awards.rb
swimsoft: When we create new transfer ex: burn the source is set by default earned. is it a bug or should it be by default earned? that's why I created the migration to fix source for all transfers.

app/controllers/dashboard/transfers_controller.rb
```diff
@@ -44,6 +47,7 @@ def query
         .includes(issuer: [image_attachment: :blob], account: [image_attachment: :blob])
 
       @transfers_unfiltered = @transfers_unfiltered.not_cancelled unless params.fetch(:q, {}).fetch(:filter, nil) == 'cancelled'
+      @transfers_unfiltered = @transfers_unfiltered.not_burned unless transfer_type_name == 'burn'
```
matjacdev515, line 50: Think we can avoid unless here by `if transfer_type_name != 'burn'`

app/controllers/dashboard/transfers_controller.rb
```diff
@@ -75,6 +79,10 @@ def set_award_type
       @award_type = @project.default_award_type
     end
 
+    def transfer_type_name
+      @transfer_type_name ||= TransferType.find_by(id: params.fetch(:q, {}).fetch(:transfer_type_id_eq, nil))&.name
```
matjacdev515, line 83: TransferType.find_by(id: params.dig(:q, :transfer_type_id_eq))&.name

app/decorators/project_decorator.rb
```diff
@@ -162,18 +162,24 @@ def transfers_chart_types
...
-    project.transfer_types.pluck(:name).reverse.map.with_index { |t, i| [t, Comakery::ChartColors.lookup(i)] }.to_h
+    project.transfer_types.pluck(:name).map.with_index { |t, i| [t, Comakery::ChartColors.lookup(i)] }.to_h
   end
 
   def transfers_chart_colors_objects
     project.transfer_types.map.with_index { |t, i| [t, Comakery::ChartColors.lookup(i)] }.to_h
   end
 
+  def transfers_chart_amount(negative, amount)
+    return -amount if negative
```
matjacdev515, line 173: Think this whole method can be replaced with simple`number.abs`

db/data_migrations/20210205131554_fix_source_for_awards.rb
matjacdev515: Can we use `update_column` instead?

app/controllers/dashboard/transfers_controller.rb
```diff
@@ -22,9 +22,12 @@ def create
     authorize @project, :create_transfer?
 
     @transfer = @award_type.awards.new(transfer_params)
-    @transfer.name = @transfer.transfer_type.name.titlecase
+    name = @transfer.transfer_type.name
+
+    @transfer.name = name.titlecase
     @transfer.account_id = params[:award][:account_id]
     @transfer.issuer = current_account
+    @transfer.source = Award.sources[name.to_sym] || 0
```
matjacdev515, line 30: You can do `Award.sources[name.to_sym].to_i`

matjacdev515 (changes requested):

app/decorators/project_decorator.rb
matjacdev515: https://ruby-doc.org/core-2.5.0/Numeric.html#method-i-abs

matjacdev515: NVM

db/data_migrations/20210205131554_fix_source_for_awards.rb
Alexey1100: Yes, source shouldn't be used anymore after we implemented the transfer types and had the migration below, but looks like there's still some code which sets the source.

https://github.com/CoMakery/comakery-server/blob/d86b5d673e8520d09632d91217cf6b0f157e6a28/db/data_migrations/20200703123332_migrate_award_source_to_transfer_type.rb#L7

app/controllers/dashboard/transfers_controller.rb
swimsoft: I used update because using **update_column** skips validations
unless is the opposite of if what's the point to avoid it?

db/data_migrations/20210205131554_fix_source_for_awards.rb
swimsoft: > Can we use `update_column` instead?

I used update because using update_column skips validations

aquabu: Probably better to include validations and also try running the migrations against demo server data locally.

matjacdev515 (approved):

aquabu (approved):

PR #1873, opened 2021-02-10
OreIdAccount: Set unclaimed state on password reset

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/176881973

### DIFF
--- a/app/controllers/api/v1/wallets_controller.rb
+++ b/app/controllers/api/v1/wallets_controller.rb
@@ -54,8 +54,10 @@ class Api::V1::WalletsController < Api::V1::ApiController
 
   # POST /api/v1/accounts/1/wallets/1/password_reset
   def password_reset
-    @auth_url = wallet.ore_id_account.service.authorization_url(redirect_url, params.dig(:proof, :signature))
-    wallet.ore_id_account.schedule_password_update_sync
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/wallets_controller.rb
```diff
@@ -54,8 +54,10 @@ def destroy
 
   # POST /api/v1/accounts/1/wallets/1/password_reset
   def password_reset
-    @auth_url = wallet.ore_id_account.service.authorization_url(redirect_url, params.dig(:proof, :signature))
-    wallet.ore_id_account.schedule_password_update_sync
+    ore_id_account = wallet.ore_id_account
+    @auth_url = ore_id_account.service.authorization_url(redirect_url, params.dig(:proof, :signature))
+    ore_id_account.unclaimed!
```
Alexey1100, line 59: I think it's better to move the state update from controller into the model, where we have all other state-related logic.

We can move it directly into `OreIdAccount#sync_password_update`

aquabu (approved):

app/controllers/api/v1/wallets_controller.rb
aquabu: possible refactor in follow on PR

PR #1874, opened 2021-02-10
Generate API docs on deployment

swimsoft:
Task: https://www.pivotaltracker.com/story/show/176726585


config.ru
```diff
@@ -1,4 +1,5 @@
 # This file is used by Rack-based servers to start the application.
 
 require ::File.expand_path('config/environment', __dir__)
+system('RAILS_ENV=test bundle exec rake docs:generate &') if Rails.env.development?
```
Riddlerrr, line 4: Is the command will be run on each call? For example, if you run `bundle exec rails console` it will generate docs before start?

app.json
```diff
@@ -5,7 +5,7 @@
   "logo": "https://avatars0.githubusercontent.com/u/17172662",
   "success_url": "/",
   "scripts": {
-    "postdeploy": "bundle exec rails whitelabel_mission:create_default"
+    "postdeploy": "bundle exec rails whitelabel_mission:create_default && bundle exec bin/docs"
```
Alexey1100, line 8: I think post deploy runs only once, when Heroku app has been deployed from the `app.json`.

We need to add the logic into `bin/after_deploy_script` when we have all the migrations, so it will work properly.

Alexey1100 (changes requested):
Should run on every deploy

bin/after_deploy_script
```diff
@@ -20,3 +20,6 @@ system! 'bin/rails db:migrate'
 
 puts "\n== Data migrations =="
 system! 'bin/rails data:migrate'
+
+puts "\n== Generate docs =="
+system! 'RAILS_ENV=test bundle exec rake docs:generate'
```
Alexey1100, line 25: 👍

Let's use bin/rails for consistency with other calls in the script
```suggestion
system! 'bin/rails docs:generate'
```

Alexey1100 (approved):

Riddlerrr (approved):

aquabu (approved):

PR #1892, opened 2021-02-17
Fast CI #1: split GitHub action workflow into parallel jobs

matjacdev515:


.github/workflows/rspec.yml
```diff
@@ -130,28 +118,13 @@ jobs:
...
-      run: bin/erblint_ci
-
-    - name: Run Brakeman
-      run: bundle exec brakeman --exit-on-warn --quiet
-
     - name: Precompile assets
       run: bundle exec rails assets:precompile
 
-    - name: Run migrations
-      run: bundle exec rails db:migrate
```
Alexey1100, line 146: I think we wanted to keep it to make sure that the db migrations are valid.

aquabu: yes, we want to keep that to make sure the migrations are valid. They have a tendency to break and then we only know when we deploy to an older environment.

matjacdev515: It is what `db:test:prepare` does

spec/rails_helper.rb
```diff
@@ -3,8 +3,8 @@
 
 require 'simplecov'
 SimpleCov.start :rails do
-  SimpleCov.minimum_coverage 98
-  SimpleCov.minimum_coverage_by_file 50
+  # SimpleCov.minimum_coverage 98
```
aquabu, line 6: Looks like minimum coverage breaks because of splitting the tests across multiple runs. The code coverage minimum is important for keeping the code quality high.

Is it possible to limit the minimum coverage match the area of the application being tested?
```
command: [
    name: Lints, Specs and Security Checks	          
          'spec/features',
          'spec/controllers',
          'spec/models',
          '--exclude-pattern "spec/{features,controllers, models}/**/*_spec.rb"'
]
```

This may not work for features but may work for models & controllers.

matjacdev515: This is a global setting for SimpleCov, I can't modify it from github actions. We can try to use env variable but it will be overcomplicated on the github actions workflow side. I commented this for now because my next task is parallel testing and parallel jobs on the github actions will be not needed. If it will work, I'll think about coverage on the parallel testing side or will back to this and try to resolve in context of github actions.

aquabu (approved):
Since the next task is to implement test parallelization for rspec, it's fine to remove the minimum test coverage and get the benefit of the test speed increase. We should just keep in mind this is off for the time being.

PR #1895, opened 2021-02-18
Bugfixes and further OreID UI integrations

Alexey1100:
- Update transfer rules, freeze, and accounts UI to support Algorand Security Token via OreID
- Fix datetime input for transfer rules and account records
- Refactor stimulus controller for accounts
- Store AccountTokenRecord wallet address in BlockchainTransaction
destination
- Fix appArgs integer encoding incompatabilities with AlgoIndexer (allow
up to 64 bit uints)
- Validate presence of address for ore_id wallets only if wallet
    provisions are present
- Fix reschedule logic for wallet provision jobs
- Add `TokenType#default_reg_group` method to security tokens
- Add `TokenType#transfer_rule_sync_job` method to security tokens
- Add `TokenType#accounts_sync_job` method to security tokens
- Implement jobs to sync AlgorandSecurityToken transfer rules and accounts with blockchain
- Add `Refreshable` concern to calculate whether the scope should be synced with blockchain at the moment
- Add Refresh button to Accounts UI for AlgorandSecurityToken

Resolves:
https://www.pivotaltracker.com/story/show/175049220
https://www.pivotaltracker.com/story/show/176909035
https://www.pivotaltracker.com/story/show/176909053
https://www.pivotaltracker.com/story/show/177045614
https://www.pivotaltracker.com/story/show/176990770
https://www.pivotaltracker.com/story/show/177050780

Known issues with transfer rules UI:
- Freeze button doesn’t display tx progress (staying in old state until tx is succeed)
- Failed transfer rules creation stays in pending state until new transfer rule with the same groups is created
- Deletion of a transfer rule doesn’t display tx progress properly (backend creates a new rule to perform the “delete”, so the rule is staying in the old state, and new rule appears in pending state until it’s confirmed, after that both rules are removed)

### DIFF
--- a/app/controllers/sign/ore_id_controller.rb
+++ b/app/controllers/sign/ore_id_controller.rb
@@ -1,14 +1,13 @@
 class Sign::OreIdController < ApplicationController
   include OreIdCallbacks
-  skip_after_action :verify_authorized, only: :receive, unless: :verify_errorless
-  skip_after_action :verify_authorized, only: :new, unless: :new_transaction_for_award?
+  skip_after_action :verify_authorized, only: %i[new receive]
 
   # POST /sign/ore_id/new
   def new
     # TODO: Add token admin role to policy token related tx
-    authorize new_transaction.blockchain_transactable, :pay? if new_transaction.is_a?(BlockchainTransactionAward)
+    authorize new_transaction.blockchain_transactable, :pay? if new_transaction_for_award?
 
-    new_transaction.source = current_account.address_for_blockchain(new_transaction.blockchain_transactable._blockchain)
+    new_transaction.source = current_account.address_for_blockchain(new_transaction.blockchain_transactable.token._blockchain)
     new_transaction.save!
 
     redirect_to sign_url(new_transaction)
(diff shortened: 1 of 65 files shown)


app/controllers/sign/ore_id_controller.rb
```diff
@@ -1,21 +1,21 @@
...
 
   # POST /sign/ore_id/new
+  # GET /sign/ore_id/new
   def new
     # TODO: Add token admin role to policy token related tx
-    authorize new_transaction.blockchain_transactable, :pay? if new_transaction.is_a?(BlockchainTransactionAward)
+    authorize new_transaction.blockchain_transactable, :pay? if new_transaction_for_award?
 
-    new_transaction.source = current_account.address_for_blockchain(new_transaction.blockchain_transactable._blockchain)
+    new_transaction.source = current_account.address_for_blockchain(new_transaction.blockchain_transactable.token._blockchain)
```
aquabu, line 11: some of this code in the new method should be broken out into separate methods that are unit tested.

Alexey1100: @noah Yeah, I think splitting `new transaction` helper will definitely improve the code, but we still have to do branching based on parameter name, since we accept a lot of different resources (transactables) here.

Currently though, the action itself is covered by extensive integration tests:
https://github.com/CoMakery/comakery-server/blob/6a76dee77ea9810974db55d14aea5d12ebd5a789/spec/controllers/sign/ore_id_controller_spec.rb#L17

aquabu (approved):

Riddlerrr (approved):
There are a lot of changes in one PR so it is hard to review in details. But overall it looks good 👍

PR #1905, opened 2021-02-25
Implement Recovery Service

Alexey1100:
- Integrate ECIES lib with additional modifications to support public
key calculation
- Add `ApiOreIdWalletRecovery` model
- Update `ApiWalletsController#password_reset` to include request
signature as a param
- Add `ApiWalletRecoveryController#recover` endpoint
- Improve spec coverage


Riddlerrr (approved):
Quickly scanned it, looks good to me 👍

db/migrate/20210224145046_create_api_ore_id_wallet_recoveries.rb
```diff
@@ -0,0 +1,9 @@
+class CreateApiOreIdWalletRecoveries < ActiveRecord::Migration[6.0]
+  def change
+    create_table :api_ore_id_wallet_recoveries do |t|
+      t.belongs_to :api_request_log, foreign_key: true, null: false, index: { unique: true }
```
aquabu, line 4: we don't want to log the payload.

spec/support/mom.rb
```diff
@@ -1163,6 +1163,18 @@ def wallet_provision(**attrs) # rubocop:todo Metrics/CyclomaticComplexity
   WalletProvision.new(params.merge(attrs))
 end
 
+def api_request_log(**attrs)
+  params = {
+    signature: 'test_signature',
+    ip: IPAddr.new('0.0.0.0'),
+    body: {
+      test: :test
```
aquabu, line 1171: don't log the payload

spec/controllers/api/v1/wallet_recovery_controller_spec.rb
```diff
@@ -13,48 +13,137 @@
...
-        expect(response.body).to eq '{"errors":{"invalid_env_variable":"WALLET_RECOVERY_WRAPPING_KEY variable was not configured or use wrong format"}}'
+      post :recover, params: params
+    end
+
+    context 'with invalid recovery_token' do
+      let(:recovery_token) { '0' }
+
+      it { is_expected.to have_http_status(:unauthorized) }
+    end
+
```
aquabu, line 103: Write a test to confirm that the payload isn't stored in the logging service. Also the payload should not be logged in the weblogs (similar to password filtering).

db/migrate/20210224145046_create_api_ore_id_wallet_recoveries.rb
Alexey1100: Yeah, there's nothing stored in database except the api_request_log_id.

Alexey1100: But, we need to filter it out from api_request_log table, since we are logging every request there.

aquabu:
Addressing requested fixes in a follow on commit.

PR #1928, opened 2021-03-11
Feature/parallelize rspec tests

matjacdev515:


knapsack_rspec_report.json
```diff
@@ -0,0 +1,277 @@
+{
```
Alexey1100, line 1: Looks like this should be in .gitignore

aquabu:
# Speed

Looks like its running at around 10m with Knapsack pro. It may rebalance tests to faster speed. If we could cache pre-compilation it would speed up by another 3 min.

# Code Coverage

Seems like it's not writing the entire coverage out to the rcov artifact file. Needs to do that and verify that test coverage is at 98%.

aquabu (changes requested):
I merged in the 15x parallelism for knapsack. We may be able to speed the build up in future commits by caching asset compilation.

To complete this PR though it needs to aggregate all the coverage data and fail if the coverage is < 98% for the full run.

aquabu:
I made the build info public to make it easier to debug: https://knapsackpro.com/dashboard/organizations/2193/projects/1396/test_suites/1992/builds

aquabu (approved):
Merging this to get the test speed improvements. We can add the test coverage in a follow on PR. I'll create a task for it.

PR #1946, opened 2021-03-19
update app.json and include AWS_REGION as ENV

shoaibmalik786:
Please set `AWS_REGION` as the ENV variables on server to make it work.

FYI - `METAMASK_LOGIN should have a default value of false and be required` - This is already as it should be in app.json

Thanks


aquabu:
@shoaibmalik786 I think the issue with MetaMask is that the default false value was a boolean instead of the string "false". When deploying to Heroku it ignored it. I modified it for testing.

aquabu (changes requested):
I'm getting this error on attempted deployment to Heroku:
Aws::Errors::MissingRegionError: No region was provided. Configure the `:region` option or export the region name to ENV['AWS_REGION']

shoaibmalik786:
@aquabu - I have updated the app.json. Please check now

Thanks

aquabu:
Hmm. Well I might have suggested something that won't work
.
The goal is to:
* Use AWS if available
* If AWS is not available, run the app using only ActiveStorage

Apparently the dummy value approach is causing an issue. 

I tried to deploy to Heroku and got:
```
Known AWS regions include (not specific to this service):
       
       af-south-1
       ap-east-1
       ap-northeast-1
       ap-northeast-2
       ap-south-1
       ap-southeast-1
       ap-southeast-2
       ca-central-1
       eu-central-1
       eu-north-1
       eu-south-1
       eu-west-1
       eu-west-2
       eu-west-3
       me-south-1
       sa-east-1
       us-east-1
       us-east-2
       us-west-1
       us-west-2
       cn-north-1
       cn-northwest-1
       us-gov-east-1
       us-gov-west-1
       us-iso-east-1
       us-isob-east-1
       /tmp/build_854fc00a/vendor/bundle/ruby/2.7.0/gems/aws-sdk-core-3.111.0/lib/aws-sdk-core/plugins/regional_endpoint.rb:42:in `block in <class:RegionalEndpoint>'
       /tmp/build_854fc00a/vendor/bundle/ruby/2.7.0/gems/aws-sdk-core-3.111.0/lib/seahorse/client/configuration.rb:72:in `call'
```

aquabu:
@shoaibmalik786 

We added some things to this and created a new PR https://github.com/CoMakery/comakery-server/pull/1959

I'm closing this one and merging https://github.com/CoMakery/comakery-server/pull/1959

Issue #1941, opened 2021-03-19
Add reCaptcha v3 to the signup and sign in pages


aquabu:
Here are the developer docs for reCaptcha v3:
https://developers.google.com/recaptcha/docs/v3

## Acceptance:

* Captcha is used on the signup page in the lower right of the page
* Captcha is used on the login page in the lower right of the page
* Any API key is stored in an environment variable - these will be set on Heroku deploys
* If the reCAPTCHA API key is not available the captcha does not appear and there are no errors
* Any API environment variable is added to the app.json as a default environment variable
* Take a screenshot of the signup and login page and add it to the pull request so we can check the appearance in advance

aquabu:
Done. Thanks https://github.com/CoMakery/comakery-server/pull/1950

PR #1950, opened 2021-03-22
Recaptcha

kolishchak:


hotwallet/.env.example
```diff
@@ -6,3 +6,5 @@ REDIS_URL=redis://localhost:6379/2
 CHECK_FOR_NEW_TRANSACTIONS_DELAY=30
 OPT_IN_APP=none
 BLOCKCHAIN_NETWORK=algorand_test
+RECAPTCHA_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
```
Riddlerrr, line 9: This file used for Hot Wallet. It's a separate service. Could you delete those ENVs from that file?

Riddlerrr (changes requested):

Riddlerrr:
@kolishchak Could you run `bin/rubocop` to fix linter issues.

spec/controllers/sessions_controller_spec.rb
```diff
@@ -148,6 +148,8 @@
     let!(:account2) { create(:account, email: 'user2@example.com', password: nil) }
     let(:project) { create(:project, account: account1, public: false, maximum_tokens: 100_000_000, token: create(:token, _token_type: 'eth', _blockchain: :ethereum_ropsten)) }
 
+    before { allow_any_instance_of(RecaptchaVerifier).to receive(:valid?).and_return(true) }
```
Alexey1100, line 151: Looks like the logic is not triggered anyway, since keys are not present in test env.

We would need to add a spec to cover RecaptchaVerifier itself, mocking the variables.

Alexey1100 (changes requested):

.env.required
```diff
@@ -9,3 +9,5 @@ SLACK_API_KEY=none
 SLACK_API_SECRET=none
 REFILE_S3_REGION=none
 REFILE_S3_BUCKET=none
+RECAPTCHA_SITE_KEY=none
```
Alexey1100, line 12: This list contains keys which are required for every environment (see `config/initializers/env_verifier.rb`), and since recaptha is not required, we should remove it from here.

We have `app.json` as a reference list of all env variables, that should be enough :)

kolishchak: Gotcha
Thanks :)

config/initializers/recaptcha.rb
```diff
@@ -0,0 +1,4 @@
+Recaptcha.configure do |config|
+  config.site_key = ENV['RECAPTCHA_SITE_KEY']
```
Alexey1100, line 2: Will the initialiser raise if the keys are empty?

spec/interactors/accounts/register_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails_helper'
+
+describe Accounts::Register do
+  describe '#call' do
```
Alexey1100, line 4: We need to cover whitelabel registration here as well, otherwise simplecov will mark the branch as uncovered, once we set it up :)

Alexey1100 (changes requested):

app/controllers/accounts_controller.rb
```diff
@@ -40,33 +40,26 @@ def show # rubocop:todo Metrics/CyclomaticComplexity
   end
 
   def create
-    @account = if @whitelabel_mission
-      @whitelabel_mission.managed_accounts.new(account_params)
-    else
-      Account.new(account_params)
-    end
+    register_account_result = Accounts::Register.call(whitelabel_mission: @whitelabel_mission,
```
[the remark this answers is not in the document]
kolishchak: @Alexey1100 we don't create an account, we only initialize it.

In my opinion it's better to provide the account with his filled data and display the error of reCaptcha verification.

spec/interactors/accounts/register_spec.rb
kolishchak: Done. Thanks for the tip :)

config/initializers/recaptcha.rb
kolishchak: No

app/controllers/accounts_controller.rb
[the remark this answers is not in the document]
Alexey1100: @kolishchak 

https://github.com/CoMakery/comakery-server/blob/1630c21c9230fe7dafab5747f4421e7a8ca8af2f/app/interactors/accounts/register.rb#L22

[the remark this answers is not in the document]
Alexey1100: Yep, I think it makes sense to initialise the instance, but currently the interactor tries to save the record.

Riddlerrr (changes requested):
Code looks good to me 👍

But when I trying to sign in locally I always getting "Invalid email or password" error
Tried to sing in on `acceptance` branch with the same credentials and it works well.
So it seems something not work properly.
If I understood correctly it's not required any new ENV for the development environment it should just skip recaptcha check, right?

Riddlerrr:
@kolishchak Could you attach a screenshot with the Recaptcha placed on a page?

zhinfuhnya:
@Riddlerrr you're right, without the keys it gives out `false` result.

I updated the reCaptcha config as follows:
https://github.com/CoMakery/comakery-server/pull/1950/files#diff-4c1c0812f6507ae00bdd0349a238b9ab95fc9240c4ce2518e5830c45431b0cdaR4

Riddlerrr:
@zhinfuhnya Now I can sign in with no Recaptcha ENVs locally but can't with configured ones. Not sure is it ok for dev env?
Probably something not configured on Recaptcha side.
Could you check it works on your side?

CoMakery - Achieve Big Missions 2021-03-26 15-01-28

zhinfuhnya:
@Riddlerrr on my host once again

image

image

P.S.: you have to register the credentials under the host, that's probably why it didn't work for you : )

Riddlerrr:
@zhinfuhnya Yeah, I thought about it too. Ok, I think we can merge and test it on demo.

I don't have comments about the code. The only comment about how it looks:
image

@aquabu I think we should do something with this ☝️

Riddlerrr (approved):

kolishchak:
@Riddlerrr the captcha is placed at the bottom right as required (although I agree that it looks terrible 🙃).

In general I don't see the point of displaying it.

@aquabu what do you think about it?
https://github.com/CoMakery/comakery-server/pull/1950#issuecomment-809498150

P.S.: reference to the ticket description ⬇️ 
https://github.com/CoMakery/comakery-server/issues/1941

aquabu:
@kolishchak if we don't have to display it, let's not display it :)

kolishchak:
@aquabu last minor style changes:
https://github.com/CoMakery/comakery-server/pull/1950/commits/b9019a57db0022df25290a3c0a0a9ac756a8c41c

All is done 🚀

Alexey1100 (approved):

PR #1956, opened 2021-03-24
Recovery token lifetime

kolishchak:
Given a 5 minute delay from the generation of the recoveryToken
When I POST /api/v1/wallet/recovery/recover with the recoveryToken
Then I get a 401 unauthorized error

## Acceptance

* update the controller code
* automated test verifying that the recoveryToken expires after 5 minutes

## Related Files

* [Api::V1::WalletRecoveryController](https://github.com/CoMakery/comakery-server/blob/444a1ede4a1268e456db40eb60d1699917dbd63d/app/controllers/api/v1/wallet_recovery_controller.rb#L1)
* [Spec](https://github.com/CoMakery/comakery-server/blob/444a1ede4a1268e456db40eb60d1699917dbd63d/spec/controllers/api/v1/wallet_recovery_controller_spec.rb)


Alexey1100 (approved):
👍

I'm leaning towards adding the validation directly to `ApiOreIdWalletRecovery` instead of putting it in more generic `ApiRequestLog`, but it's definitely not blocking :)

app/models/api_request_log.rb
```diff
@@ -1,6 +1,10 @@
 class ApiRequestLog < ApplicationRecord
+  has_one :api_ore_id_wallet_recovery, dependent: :destroy
+
   validates :ip, :body, presence: true
   validates :signature, presence: true, uniqueness: true
 
-  has_one :api_ore_id_wallet_recovery, dependent: :destroy
+  def token_expired?
+    DateTime.current > created_at + 5.minutes
```
aquabu, line 8: I think the 5.minutes should be refactored out into a named constant like `ApiRequestLog::TOKEN_EXPIRATION_SECONDS=5.minutes`

aquabu (changes requested):
Minor refactor suggested.

aquabu (approved):

aquabu:
thanks for the refactor @kolishchak 👍

PR #1958, opened 2021-03-25
Bump mimemagic gem

Riddlerrr:
This PR fixes `bundle install`

There is crazy issue with `mimemagic` gem related to licencing. They removed some versions of the gem and now it requires update the gem version 🤯
It broke all rails installations because `mimemagic` is required for `activestorage`

Details: https://github.com/rails/rails/issues/41750


aquabu:
@Riddlerrr thank you. I saw the top of the rabbit hole. Glad you figured that out :)

aquabu (approved):

PR #1963, opened 2021-03-26
Transfer cancel link should not appear when the transfer is pending

shoaibmalik786:


aquabu (changes requested):
Please add rspec test coverage for each conditional branch.

shoaibmalik786:
@aquabu I have written the rspec. Please have a look.

app/policies/award_policy.rb
```diff
@@ -57,7 +57,18 @@ def pay?
   end
 
   def cancel?
-    cancel_by_contributor? || cancel_by_admin?
+    (cancel_by_contributor? || cancel_by_admin?) && status_for_cancel?
+  end
+
+  def status_for_cancel?
```
aquabu, line 63: The case statement is a convoluted which will make it difficult to maintain. There are also unnecessary code branches that could cause unexpected behavior.

I would recommend refactoring `status_for_cancel?` into a `BlockchainTransactable#cancelable?` method which has object access to `status` and `waiting_in_created?`

That method should have a unit test that covers each branch in the conditional statement.



The

aquabu (changes requested):
Thanks for the additional test. Please refactor the `status_for_cancel?` method.

app/policies/award_policy.rb
shoaibmalik786: I have made the changes.

aquabu (approved):

app/models/concerns/blockchain_transactable.rb
```diff
@@ -67,5 +67,17 @@ def new_blockchain_transaction(params)
         )
       )
     end
+
+    def cancelable?
+      flag = case latest_blockchain_transaction&.status
```
aquabu, line 72: @shoaibmalik786 thanks for the refactor. Last minor detail. The case statement can just return directly from `cancelable?`. The `flag = ` and `flag` at the end of the method aren't needed. Thanks!

shoaibmalik786: I have removed flag.

aquabu: Awesome. Thanks!

app/models/concerns/blockchain_transactable.rb
```diff
@@ -67,5 +67,16 @@ def new_blockchain_transaction(params)
         )
       )
     end
+
+    def cancelable?
+      case latest_blockchain_transaction&.status
```
Alexey1100, line 72: Seems like this doesn't cover all the statuses properly.

Transactable should be cancelable **only** if the latest transaction is created, cancelled or failed, or there are no transactions at all.

Alexey1100 (changes requested):

app/models/concerns/blockchain_transactable.rb
aquabu: @Alexey1100 hmm... ok. Let's spec that out in pivotal tracker.

PR #1969, opened 2021-03-29
Recovery service: Use wallet_recovery_api_public_key to verify a signature

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/177168118


app/controllers/api/v1/concerns/requires_recovery_signature.rb
```diff
@@ -0,0 +1,36 @@
...
+    before_action :verify_signature
+
+    def verify_signature
+      signature_verification_result =
+        Comakery::APISignature.new(
+          params.as_json,
+          request.base_url + request.path,
+          request.method,
+          ->(nonce) { nonce_unique?(nonce) }
+        ).verify(whitelabel_mission&.wallet_recovery_api_public_key)
```
Riddlerrr, line 15: This is full copy of `RequiresSignature` concern except getting `wallet_recovery_api_public_key` instead of `whitelabel_api_public_key`

Can be DRYed later

app/controllers/missions_controller.rb
```diff
@@ -48,7 +48,7 @@ def edit
 
   def update
     authorize @mission
-    if @mission.update(mission_update_params)
```
Riddlerrr, line 51: This allows us to fill the `wallet_recovery_api_public_key` by editing a mission if it is not set on creation.

app/models/mission.rb
```diff
@@ -83,8 +81,9 @@ def set_whitelabel_for_projects
       projects.update(whitelabel: whitelabel)
     end
 
-    def whitelabel_api_public_key_cannot_be_overwritten
-      errors.add(:whitelabel_api_public_key, 'cannot be overwritten')
+    def whitelabel_api_public_keys_cannot_be_overwritten
+      errors.add(:whitelabel_api_public_key, 'cannot be overwritten') if whitelabel_api_public_key_changed? && whitelabel_api_public_key_was.present?
+      errors.add(:wallet_recovery_api_public_key, 'cannot be overwritten') if wallet_recovery_api_public_key_changed? && wallet_recovery_api_public_key_was.present?
```
Riddlerrr, line 86: Do not allow editing this field

aquabu (approved):

PR #1974, opened 2021-03-30
changed redirect from root path to password recovery page

o-bahriy:
https://www.pivotaltracker.com/story/show/177124408


o-bahriy:
this won't fix update password notification(same problem with layouts)
plus i'm not sure about message, probably should change it too(all lowercased)

aquabu (approved):

PR #1984, opened 2021-04-01
WIP: Add ability to sync balance from blockchains

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/177309195

### DIFF
--- a/app/jobs/sync_balance_job.rb
+++ b/app/jobs/sync_balance_job.rb
@@ -1,0 +1,10 @@
+class SyncBalanceJob < ApplicationJob
+  queue_as :low
+
+  def perform(balance)
+    return unless balance
+    return if balance.updated_at < 10.seconds.ago
+
+    balance.sync_with_blockchain!
+  end
+end
--- a/app/lib/comakery/algorand.rb
+++ b/app/lib/comakery/algorand.rb
@@ -58,7 +58,7 @@ class Comakery::Algorand
   end
 
   def account_balance(addr)
-    account_details(addr).dig('account', 'amount')
+    account_details(addr).dig('account', 'amount') || 0
   end
 
   def account_assets(addr)
@@ -75,6 +75,14 @@ class Comakery::Algorand
 
   def account_local_state_decoded(addr)
     decode_storage(account_local_state(addr)['key-value'])
+  end
+
+  def app_balance(addr)
+    account_local_state_decoded(addr).fetch('balance', 0)
+  end
+
+  def asset_balance(addr)
+    account_assets(addr).find { |asset| asset['asset-id'] == @asset_id }&.fetch('amount', 0)
   end
 
   def status
--- a/app/models/account_token_record.rb
+++ b/app/models/account_token_record.rb
@@ -20,7 +20,6 @@ class AccountTokenRecord < ApplicationRecord
   attr_readonly :account_id, :token_id, :reg_group_id, :max_balance, :account_frozen, :lockup_until, :balance
   validates_with SecurityTokenValidator
   validates :lockup_until, inclusion: { in: LOCKUP_UNTIL_MIN..LOCKUP_UNTIL_MAX }
-  validates :balance, inclusion: { in: BALANCE_MIN..BALANCE_MAX }, allow_nil: true
   validates :max_balance, inclusion: { in: BALANCE_MIN..BALANCE_MAX }
 
   enum status: { created: 0, pending: 1, synced: 2, failed: 3 }
@@ -34,6 +33,12 @@ class AccountTokenRecord < ApplicationRecord
 
   def lockup_until=(time)
     super(time.to_i.to_d)
+  end
+
+  def balance
+    balance = Balance.find_or_create_by(wallet_id: wallet_id, token_id: token_id)
+    SyncBalanceJob.set(queue: :critical).perform_later(balance)
+    balance
   end
 
   private
--- a/app/models/balance.rb
+++ b/app/models/balance.rb
@@ -9,4 +9,12 @@ class Balance < ApplicationRecord
   def value
     token.from_base_unit(base_unit_value)
   end
+
+  def blockchain_balance_value
+    @blockchain_balance ||= token.blockchain_balance(wallet.address)
+  end
+
+  def sync_with_blockchain!
+    update(base_unit_value: blockchain_balance_value)
+  end
 end
--- a/app/models/token_type.rb
+++ b/app/models/token_type.rb
@@ -17,4 +17,10 @@ class TokenType
   def self.append_to_list(token_type)
     list.merge(token_type => (list.values.max + 1))
   end
+
+  # Return balance of symbol for provided addr
+  # Should be implemented for every particular token type
+  def blockchain_balance(_wallet_address)
+    raise NotImplementedError
+  end
 end
--- a/app/models/token.rb
+++ b/app/models/token.rb
@@ -119,6 +119,10 @@ class Token < ApplicationRecord
     RegGroup.default_for(self)
   end
 
+  def blockchain_balance(wallet_address)
+    token_type.blockchain_balance(wallet_address)
+  end
+
   private
 
     def set_values_from_token_type # rubocop:todo Metrics/CyclomaticComplexity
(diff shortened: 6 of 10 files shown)


app/lib/comakery/algorand.rb
```diff
@@ -77,6 +77,14 @@ def account_local_state_decoded(addr)
     decode_storage(account_local_state(addr)['key-value'])
   end
 
+  def app_balance(addr)
+    account_local_state_decoded(addr).fetch('balance', 0)
+  end
+
+  def asset_balance(addr)
+    account_assets(addr).find { |asset| asset['asset-id'] == @asset_id }&.fetch('amount', 0)
```
Alexey1100, line 85: This might return nil

app/models/token_type.rb
```diff
@@ -17,4 +17,10 @@ def self.list
   def self.append_to_list(token_type)
     list.merge(token_type => (list.values.max + 1))
   end
+
+  # Return balance of symbol for provided addr
```
Alexey1100, line 21: All of the instance methods should go into token type template and all of the subclasses.
That was our initial design, to make subclasses more readable.

db/schema.rb
```diff
@@ -263,7 +263,7 @@
   create_table "balances", force: :cascade do |t|
     t.bigint "wallet_id"
     t.bigint "token_id"
-    t.bigint "base_unit_value", default: 0, null: false
+    t.decimal "base_unit_value", precision: 40, default: "0", null: false
```
Alexey1100, line 266: If using decimal we need to make the precision 78 to fit uint256.

But we don't need decimal here, since base unit is always integer.

zhinfuhnya (approved):

app/models/token_type/comakery_security_token.rb
```diff
@@ -70,6 +70,8 @@ def accounts_sync_job
     ::BlockchainJob::ComakerySecurityTokenJob::AccountTokenRecordsSyncJob
   end
 
+  # Return balance of symbol for provided addr
+  # @return [Integer] balance
   def blockchain_balance(wallet_address)
     contract.contract.call.balance_of(wallet_address)
```
Alexey1100, line 76: `contract` method is implemented in `TokenType::Erc20`.
So yeah, we can move the balance there easily.

app/jobs/sync_balance_job.rb
```diff
@@ -0,0 +1,11 @@
+class SyncBalanceJob < ApplicationJob
+  queue_as :low
+
+  def perform(balance)
+    return unless balance
+    # Do not update if it was updated recently but should be updated for just created balance
+    return if balance.updated_at > balance.created_at && balance.updated_at > 10.seconds.ago
```
aquabu, line 7: should `balance.updated_at > 10.seconds.ago` be reversed to `balance.updated_at < 10.seconds.ago`

app/models/account_token_record.rb
```diff
@@ -36,6 +35,12 @@ def lockup_until=(time)
     super(time.to_i.to_d)
   end
 
+  def balance
+    balance = Balance.find_or_create_by(wallet_id: wallet_id, token_id: token_id)
+    SyncBalanceJob.set(queue: :critical).perform_later(balance)
```
aquabu, line 40: why is this one `queue: critical`?

app/jobs/sync_balance_job.rb
Riddlerrr: nope, for example let's say the balance was updated 15 seconds ago:
```
[8] pry(main)> 15.seconds.ago > 10.seconds.ago
=> false
```

This means we will NOT run early return, so we WILL sync the balance.

app/models/account_token_record.rb
Riddlerrr: If the question about naming, it's one of the default queue names:
https://github.com/CoMakery/comakery-server/blob/a55a95b8055fdc33388a0d4a9f4710263c4650e7/config/sidekiq.yml#L5

About why we need this. It's one of the requirements from the task =)

> 1. If you request the balance from the database model it will also kick off a medium/high priority background job for that account so that if it's old it will update the balance. It will be available after a second call within a second or so. This forces the update.

app/jobs/sync_balance_job.rb
aquabu: Got it. Maybe better to avoid the double negative.

Might be clearer to do something like this on the Balance class:

```ruby
class Balance
  def ready_for_update?
    updated_at > 10.seconds.ago || updated_at == created_at
  end
end
```

Then the check in this SyncBalanceJob would be:
```ruby
balance.sync_with_blockchain! if balance&.ready_for_update?
```

Riddlerrr: Good idea

app/models/balance.rb
```diff
@@ -6,7 +6,22 @@ class Balance < ApplicationRecord
   validates :wallet_id, uniqueness: { scope: :token_id }
   validates :token_id, inclusion: { in: ->(b) { Array(b.wallet&.tokens_of_the_blockchain&.pluck(:id)) } }
 
+  scope :ready_for_balance_update, -> { where('updated_at < ?', 10.seconds.ago) }
```
Alexey1100, line 9: Looks like we have two different conditions for the same method name.

app/models/balance.rb
```diff
@@ -6,7 +6,22 @@ class Balance < ApplicationRecord
...
+    @blockchain_balance_base_unit_value ||= token.blockchain_balance(wallet.address)
+  end
+
+  def sync_with_blockchain!
+    update(base_unit_value: blockchain_balance_base_unit_value)
+  end
+
+  # Do not update the balance if it was updated recently but should be updated for just created balance
+  def ready_for_balance_update?
+    updated_at < 10.seconds.ago || updated_at == created_at
```
Alexey1100, line 25: Looks like we have two different conditions for the same method name.

spec/jobs/sync_balance_job_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  subject { described_class.perform_now(balance) }
+
+  it 'succesfully update a value' do
+    allow_any_instance_of(Balance).to receive(:blockchain_balance_base_unit_value).and_return(999)
+    is_expected.to be true
+    expect(balance.reload.base_unit_value).to eq 999
+  end
+
+  context 'do not update a value' do
+    let(:balance) { create(:balance, created_at: 1.day.ago, base_unit_value: 10) }
```
Alexey1100, line 14: Maybe a typo here :)
We are setting created_at but the logic relies on updated_at.

app/models/balance.rb
Riddlerrr: Yes, but one with `?` in the end =)

spec/jobs/sync_balance_job_spec.rb
Riddlerrr: no typo here =)
we set `created_at`, it means it was created yesterday but just updated.
Probably will be good to explicitly set updated_at to `Time.zone.now`.

app/models/balance.rb
Alexey1100: I think making them return the same result will help with unexpected inconsistencies

Riddlerrr: One is a model method but the second is an instance method.
Looks like I didn't understand your advice =)

spec/jobs/sync_balance_job_spec.rb
Alexey1100: Ahh okay, yeah I think the more explicit the easier to understand what's happening.

Alexey1100: Definitely not critical, but I've noticed that all these tests rely on the time of execution and the fact that there would be no more that 10 seconds between the creation of the record and the assertion, which might be a point of failure if we hit some big performance issues on CI.
It's not a problem currently but it might rise up if we update the balance caching from 10 to 1-2 seconds.

Riddlerrr: Yeah, we can freeze time for that tests. I'll add it

app/models/balance.rb
Alexey1100: I think the problem is that in some cases we might get the following situation:

`balance.ready_for_balance_update? # => true`
`Balance.ready_for_balance_update.includes?(balance) # => false`

Riddlerrr: Ok, I fixed the condition for the scope. I hope I understood it correctly =)

Alexey1100 (approved):

aquabu (approved):

PR #1986, opened 2021-04-01
version upgrade to 6.0.3.5

olehyev138:
version upgrade from 6.0.3.2 to 6.0.3.5


app/controllers/accounts_controller.rb
```diff
@@ -40,24 +40,33 @@ def show # rubocop:todo Metrics/CyclomaticComplexity
   end
 
   def create
-    @account = Accounts::Initialize.call(
```
zhinfuhnya, line 43: @olehyev138 are you sure we should roll back reCaptcha changes?

https://github.com/CoMakery/comakery-server/pull/1950

aquabu:
@olehyev138 please check if the branch needs to be updated. There have been a lot of recent changes on `acceptance`

app/controllers/accounts_controller.rb
olehyev138: I made the branch a few days ago and so never updated some data from origin branch

olehyev138: I mainly focus to upgrade Rails version

PR #1992, opened 2021-04-03
Add ChainJS-compatible transaction encoding to eth tx wrappers

Alexey1100:
Resolves:
https://www.pivotaltracker.com/story/show/177502244
https://www.pivotaltracker.com/story/show/177615677

- Add object conversion to eth and erc20 tx classes
- Use object conversion in blockchain tx classes to populate tx_raw
- Trim comakery security token abi to include only functions used by Comakery
- Sign eth transactions via OREID UI
- Disable return_signed_transaction for create_tx_params


app/lib/comakery/eth/tx/erc20/mint.rb
```diff
@@ -2,4 +2,8 @@ class Comakery::Eth::Tx::Erc20::Mint < Comakery::Eth::Tx::Erc20::Transfer
   def method_id
     '40c10f19'
   end
+
+  def method_name
+    'mint'
```
aquabu, line 7: I think the method_id should be generated using a hash of the function signature instead of hard coded - that would be better for auditing that the expected method is being called.

Alexey1100: Definitely.
We still need to define function input types to calculate the hash, so the types would be hardcoded, or we can try to extract the id from eth lib, which already parses the abi.

app/lib/comakery/eth/tx/erc20/security_token/pause.rb
```diff
@@ -0,0 +1,19 @@
+class Comakery::Eth::Tx::Erc20::SecurityToken::Pause < Comakery::Eth::Tx::Erc20
+  def method_id
+    '8456cb59'
```
aquabu, line 3: Same comment. I think we should show the inputs used to generate the method_id hash. We are defining the arguments and function name defined here. We should provide a function signature string like `baz(uint32,bool)` that it's generated from.

Here's a link to the argument encoding docs for convenience: https://docs.soliditylang.org/en/v0.8.3/abi-spec.html?highlight=argument%20encoding#function-selector-and-argument-encoding

app/lib/comakery/eth/tx/erc20/security_token/pause.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def method_name
```
aquabu, line 6: maybe we should derive the `method_name` from the signature string... e.g. `baz(uint32,bool)` return s the `method_name` `baz`

aquabu (changes requested):
check the method_id comments

app/lib/comakery/eth/tx/erc20.rb
```diff
@@ -3,6 +3,37 @@ def method_id
...
+      contract: {
+        abi: blockchain_transaction.token.abi,
+        method: method_name,
+        parameters: encode_method_params
+      }
+    })
+  end
+
+  def method_name
+    ''
```
aquabu, line 15: To reduce risk of signing an undefined transaction I think the base class `method_id` and `method_name` should be defined as `raise NotImplementedError`.  

Since this is a template class it's not informative enough to rely on duck typing. It's also not safe to rely on the current default implementation because someone could sign a transaction for the `method_name` or `''`. This might fail, but our faith in that failure relies on some downstream library that we hope would fail for any arbitrary scenario. This could lead to an exploit or simple loss of funds such as batch processing transactions that do nothing with a high gas cost.

### Reasoning for using `NotImplementedError` instead of `NoMethodError`

There is some disagreement on which error should be raised for template interfaces. But I think we should use `raise NotImplementedError`. Here's why...

Technically [the `NotImplementedError` docs](https://ruby-doc.org/core-2.5.0/NotImplementedError.html) say it is for "when a feature is not implemented on the current platform. For example, methods depending on the fsync or fork system calls..."

This might lead us to use `NoMethodError` which [the docs say](https://ruby-doc.org/core-2.5.0/NoMethodError.html) is "Raised when a method is called on a receiver which doesn't have it defined..."

However the interface testing patter in rspec for templates and interfaces typically rely on calling `respond_to?(:method_name)`. This is also the approach advocated by Sandi Metz in [Practical Object Oriented Design in Ruby](https://www.informit.com/store/practical-object-oriented-design-an-agile-primer-using-9780134456478?ranMID=24808).

To use this testing pattern you must use `NotImplementedError` to get the desired `respond_to?` output because:
| Error | `respond_to?(:method_name)` |
| -- | -- |
| NotImplementedError | `false` (what we need for testing the interface) |
| NoMethodError | `true` (won't allow us to test for the implementation) |

Therefore we should use `NotImplementedError` to define the template clearly, avoid the dangers of generic definition, and efficiently test that the templates are fully defined.

aquabu (approved):
@Alexey1100 I'm merging this and I added a refactoring story to https://www.pivotaltracker.com/n/projects/2167210/stories/177618915

app/lib/comakery/eth/tx/erc20.rb
Alexey1100: Yep I agree, this template class needs some refactoring.

> It's also not safe to rely on the current default implementation because someone could sign a transaction for the method_name or ''.
There's no security risk though, since this template class is never initialised on it's own and the decision of what tx class should be initialised to generate or verify the transaction, placed in the BlockchainTransaction model subclasses logic, where it's hardcoded.

PR #1996, opened 2021-04-05
Avoid large file uploads for ActiveStorage and ImageMagick

zhinfuhnya:
[#176769081](https://www.pivotaltracker.com/n/projects/2167210/stories/176769081)

List of updates:
- Fixed form validation on the frontend (in case of errors a request was sent)
- Added error list display, in case of more than one
- Added frontend validation to check image size
- Added AR validation for all entities (account, token, project, award, award types, mission)
- Added `ImagePixelValidator` service (the check was made from the controller because it is impossible to pull the image from blob before validation, for more details: https://github.com/rails/rails/pull/37005/files)
- Added `CreateAward` to avoid code complexity


app/models/concerns/active_storage_validator.rb
```diff
@@ -5,7 +5,7 @@ module ActiveStorageValidator
     def validate_image_attached(*fields)
       fields.each do |field|
         validates field, content_type: %w[image/png image/jpg image/jpeg],
-                         size: { less_than: 10.megabytes, message: 'is not given between size' },
```
Alexey1100, line 8: The code looks good, but there are might be three issues to check.

1. It might be still possible to upload a non-image by forging content-type (since dimension are also applicable for videos) and cause page render to crash where we are trying to use resized image. Solution might be trying to do the resize during the validation and adding an error if an exception is raised.

2. It's still possible to upload a file to server, since the validation happens after the file is uploaded. Maybe it's possible to configure heroku client_max_body_size similar to nginx.

3. Have you checked if the flood pixel attack mentioned in ticket is reproducible on Heroku? 

@zhinfuhnya pls let me know what do you think.

zhinfuhnya: @Alexey1100 took into account your comments. Very appropriate. Thanks :)
I will explore all sides more deeply to give you an answer.

Riddlerrr (approved):
👍 Great work
I looked through the code and it looks very good

zhinfuhnya:
> 👍 Great work
> I looked through the code and it looks very good

thanks :]

Alexey1100 (commented):
@zhinfuhnya Great job!

One question, could we implement `ImagePixelValidator` as a validator concern on model level?
So we don't have to remember to include and test it for every controller action.

As it says in the docs, it might be possible to pull the image via `ActiveRecord::Base#attachment_changes[]`

zhinfuhnya:
> @zhinfuhnya Great job!
> 
> One question, could we implement `ImagePixelValidator` as a validator concern on model level?
> So we don't have to remember to include and test it for every controller action.
> 
> As it says in the docs, it might be possible to pull the image via `ActiveRecord::Base#attachment_changes[]`

Unfortunately, because of the imperfection of active storage, we cannot use this method because the image is validated first — in this step it tries to collect metadata and load it into the blob.

In the case of an image with a pixel bomb — at the moment of pixel analysis we get a timeout, i.e. we cannot reach the `before_save` callback (the level where we get access to the image).

Alexey1100:
@zhinfuhnya ahh okay.
I was just thinking how can we protect ourselves from accidently passing an invalidated image from controller.

Maybe a a model instance variable which is set by ImageValidator after validation and confirmed by model before save if attachment has changed. Need to think about that :)

app/controllers/awards_controller.rb
```diff
@@ -53,17 +53,16 @@ def edit
   end
 
   def create
-    @award = @award_type.awards.new(award_params)
-    @award.issuer = current_account
+    result = CreateAward.call(
```
aquabu, line 56: does this do the policy check? `policy(source_award.project).edit?`

app/interactors/create_award.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  private
+
+    def attach_image_from_source(award)
+      return unless context.image_from_id
+
+      source_award = Award.find(context.image_from_id.to_i)
+
+      if AwardPolicy.new(context.account, source_award).edit?
```
aquabu, line 23: Is there a spec checking that the policy is enforced?

zhinfuhnya: @aquabu I would like to clarify this logic, in the code as it was before:

`if !@award.image && params[:task][:image_from_id]
      source_award = Award.find(params[:task][:image_from_id].to_i)
      if policy(source_award.project).edit?
        @award.image.attach(source_award.image.blob) if source_award.image.attached?
      end
    end`
    
we check policy through the source aaward, in my opinion there is a mistake here because we are changing the current `@award`

kolishchak: @aquabu the specs have been updated regarding the verification of the policy.

https://github.com/CoMakery/comakery-server/pull/1996/commits/89cb039dacc98bd3dece1511857920f4835c0851

zhinfuhnya: @aquabu the specs have been updated regarding the verification of the policy.

https://github.com/CoMakery/comakery-server/pull/1996/commits/89cb039dacc98bd3dece1511857920f4835c0851

aquabu (approved):

PR #1998, opened 2021-04-06
logo URL fix

olehyev138:
Resolve:
https://www.pivotaltracker.com/n/projects/2167210/stories/177321303

Detail:
 * Current GET /api/v1/tokens returns a file link instead of a URL link with the token logo image url:

a spec verifying that Rails.application.routes.url_helpers.url_for(logo_image) as used in the API returns a full path to the image that can be used by clients of the API

### DIFF
--- a/app/decorators/token_decorator.rb
+++ b/app/decorators/token_decorator.rb
@@ -23,11 +23,7 @@ class TokenDecorator < Draper::Decorator
     _blockchain
   end
 
-  def logo_url(size = 100)
-    GetImageVariantPath.call(
-      attachment: logo_image,
-      resize_to_fill: [size, size],
-      fallback: helpers.image_url('default_project.jpg')
-    ).path
+  def logo_url
+    Rails.application.routes.url_helpers.url_for(logo_image)
   end
 end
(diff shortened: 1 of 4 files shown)


Alexey1100 (changes requested):
@olehyev138 could you pls reference the task on tracker and add a spec.

aquabu:
@olehyev138 merging. But could you write a spec for this in a follow up PR? I don't want this to break again in the future without us knowing.

aquabu (approved):

aquabu:
@olehyev138 looks like @Alexey1100 also requested the spec. Yes, please add that and then we will merge it.

olehyev138:
> @olehyev138 looks like @Alexey1100 also requested the spec. Yes, please add that and then we will merge it.

Which spec do you mean? task description from tracker?

aquabu:
@olehyev138 a spec verifying that `Rails.application.routes.url_helpers.url_for(logo_image)` returns a full path to the image that can be used by clients of the API.

Alexey1100 (approved):

aquabu (approved):

Issue #1949, opened 2021-03-21
Api doc generator outputs the same docs each time and they are checked in to the GitHub repository


aquabu:
# Problem

We need to publish the documentation generated by Rspec Rails Generator. There are several sub problems:
* It is easiest if the docs are stored in the GitHub repo with the code
* Generated code has random values which creates lots of new github commits when the API hasn't changed. This creates merge conflicts that slow down merges often by days. This is why we stopped committing docs.
* Because Rspec generates the docs it isn't appropriate to generate the docs on deployment in production or staging environments because this would require having Rspec installed and running tests in the production environment - not good.
* If we published docs from GitHub actions we would have a version synchronization problem between the version of the docs and different environments (staging, production, etc).

# Solution

* Remove the gitingore and check docs into the repository
* Implement an easy to replicate pattern to use fixed values that do not vary between test runs for random test values like user IDs.
* Modify all the API tests so that the docs do not use random values that are generated on each test run
* After running bin/docs multiple times the docs generated should be completely the same (we may write a CI test for this in a future story)
* The docs should be available from the rails server at the `/doc/api/v1` route

aquabu:
@shoaibmalik786 

This approach may work for the most common things that change between runs of generating the docs.

The timestamps can be made consistent with something like:
```Timecop.freeze(Date.today + 30) do
  assert joe.mortgage_due?
end
```

etags may be based on the timestamps

the other values can be passed into the create function of the models or mocked

aquabu:
@shoaibmalik786 just merged this one!

PR #1999, opened 2021-04-06
API Docs Changes

shoaibmalik786:


Alexey1100 (changes requested):
@shoaibmalik786 wow that's a lot of work, but I don't think that's what we wanted.
We just wanted to predefine all ids and timestamps of resources used in the spec as described in #1949, not the complete API responses.

That will give us lots of troubles maintaining that : )

shoaibmalik786:
> @shoaibmalik786 wow that's a lot of work, but I don't think that's what we wanted.
> We just wanted to predefine all ids and timestamps of resources used in the spec as described in #1949, not the complete API responses.
> 
> That will give us lots of troubles maintaining that : )

There are so many places where we are using random generated hash as token or value and also using random generated emails in request_body, request_query_parameters, response_headers and response_body. What about that?

Should i only change the dynamic value rather than whole request and response parameter's values? Please suggest.

Alexey1100:
@shoaibmalik786 yep, all of these random values could be bypassed by passing params on create:

```ruby
create(:account, id: 55, email: 'email55@example.org')
```

Also you could create a set of helpers inside `mom.rb` which will always use only predefined values, and use these helpers for the API specs.

shoaibmalik786:
> @shoaibmalik786 yep, all of these random values could be bypassed by passing params on create:
> 
> ```ruby
> create(:account, id: 55, email: 'email55@example.org')
> ```
> 
> Also you could create a set of helpers inside `mom.rb` which will always use only predefined values, and use these helpers for the API specs.

@Alexey1100 Thanks for suggestion but i think we should use these static value in response because it will create the problem if record is already created with id 55 or something else.

Alexey1100:
@shoaibmalik786 the database is cleared for every example, so we shouldn't have the conflicts.

Alexey1100:
I think the static responses used to render the documentation will defeat the purpose of these specs :)

app/lib/comakery/api_signature.rb
```diff
@@ -108,11 +108,15 @@ def initialize(request, http_url = '', http_method = '', is_nonce_unique = ->(no
     # @param private_key [String] 64-bit keypair strict-encoded in base64
     #
     # @return [Hash] signed request
-    def sign(private_key)
```
Riddlerrr, line 111: I'd prefer do not change the main lib code for tests purposes only.
To freeze time you can use `Timecop` gem. https://github.com/travisjeffery/timecop
Like this:
```
  before do
    Timecop.freeze(Time.local(1990))
  end

  after do
    Timecop.return
  end
```
`nonce` also can be stubbed using RSpec:
```
allow_any_instance_of(Comakery::APISignature).to receive(:nonce).and_return('0242d70898bcf3fbb5fa334d1d87804f')
```

shoaibmalik786: Ok Thanks

aquabu:
@shoaibmalik786 @Alexey1100 @Riddlerrr  what's the status on this one? It looks like there are some merge conflicts and failing specs - but other than that how's it going? Would be great to get this one across the finish line. Thanks!

shoaibmalik786:
@aquabu I have completed the required changes for this task. some spec are falling here, not able to find the issue because on my machine all specs working fine.

aquabu:
@shoaibmalik786 there is some CI instability that has been mostly addressed on `acceptance`. I reran the checks here and they passed.

aquabu (changes requested):
@shoaibmalik786 can you remove the schema.rb changes. There shouldn't be any migrations or changes and it does some odd things like not including indexes and removing tables. If CI is failing try re-running it.

spec/acceptance/api/v1/11_general_spec.rb
```diff
@@ -36,7 +52,9 @@
         ].join(' ')
 
         request = build(:api_signed_request, '', api_v1_projects_path, 'GET', 'example.org')
-        do_request(request)
+        result = do_request(request)
+
+        result[0][:response_headers]['ETag'] = 'ETag: W/"65619e25d426a61fdaaea38f54f63b1f"' if status == 200
```
aquabu, line 57: @shoaibmalik786 we want to avoid the pattern of stubbing every ETag.

Here are a few approaches that might work in order of preference:
* If all the attributes in the response are static (as they should be from the other test changes), the ETag should stay the same and the ETag won't need to be stubbed.
* If the ETag is still changing then the ETag output could be stubbed in the controller, response object or the class generating the etag. This stub could be defined in a test helper that's reusable for all rspec tests that generate api docs. This would keep us from needing to stub the value many times in tests and make the approach easily reusable across tests.
* Finally the Etag could be stripped off the request and only be specified on the authentication pages.

aquabu:
@shoaibmalik786 thanks for resolving the merge conflicts. 

There are two other items to resolve before merging:
* remove changes to `schema.rb`
* change the way that ETags are stubbed

There are more details on these above.

shoaibmalik786:
> @shoaibmalik786 thanks for resolving the merge conflicts.
> 
> There are two other items to resolve before merging:
> 
> * remove changes to `schema.rb`
> * change the way that ETags are stubbed
> 
> There are more details on these above.

I have made the changes. There were some conflicts in schema file which i have resolved. Etag issue also fixed. there are only one spec where i have manually set the Etag value otherwise for all specs Etag value is same every time.

aquabu:
@shoaibmalik786 if you could, please remove all changes to the schema.rb file so it doesn't appear in the diff. It's hard to tell what's a change vs a reordering and there shouldn't be any schema changes needed. Thanks!

shoaibmalik786:
> @shoaibmalik786 if you could, please remove all changes to the schema.rb file so it doesn't appear in the diff. It's hard to tell what's a change vs a reordering and there shouldn't be any schema changes needed. Thanks!

@aquabu I have reverted schema file.

aquabu (approved):

PR #2000, opened 2021-04-06
fix wait index page

olehyev138:
resolve:
https://www.pivotaltracker.com/n/projects/2167210/stories/177606073


aquabu (approved):

Alexey1100:
@o-bahriy what's the reason here changing all `console.error` to `console.log`?
We do want to use console errors.

PR #2005, opened 2021-04-07
Add a load test for sidekiq jobs

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/177660832

It can be ran from a console: `LoadTest::Sidekiq.perform(10,10)`
Params:
`perform(numder_of_jobs, sleep_for_each_job)`

### DIFF
--- a/app/lib/load_test/sidekiq.rb
+++ b/app/lib/load_test/sidekiq.rb
@@ -11,8 +11,11 @@ module LoadTest
   end
 
   class LoadTestJob < ApplicationJob
+    retry_on StandardError, wait: :exponentially_longer, attempts: 3
+
     def perform(sleep_time)
       sleep sleep_time
+      raise StandardError if rand(25).zero?
     end
   end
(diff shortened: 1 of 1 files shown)


Alexey1100 (approved):
👍

I think it would be nice to throw in some random failures as well.

```ruby
  class LoadTestJob < ApplicationJob
     retry_on StandardError, wait: :exponentially_longer, attempts: 3
     
     def perform(sleep_time)
       sleep sleep_time
       raise StandardError if rand(25) == 0
     end
   end
```

Riddlerrr:
@Alexey1100 Great idea!

aquabu:
@Riddlerrr @Riddlerrr Cool. Could you make the random failures a parameter?

It would be helpful to be able to create the number of failures that background jobs will have. This will allow us to test both predictable load and random failures.

https://www.pivotaltracker.com/n/projects/2167210/stories/177672233

PR #2017, opened 2021-04-09
CI stability improvements

Riddlerrr:
- Improve CI stability
- Fix docs order


Riddlerrr:
I've tried to re-run all tests 3 times. All of those passed, so it seems pretty stable.

Riddlerrr:
For now I upload 10 simplecov reports as artifacts.
We need to find a way how to merge it all together

.github/workflows/rspec.yml
```diff
@@ -18,6 +18,7 @@ jobs:
   rspec:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
aquabu, line 21: @Riddlerrr won't this make it take longer for us to know that CI failed? Or is this so we can see all failures at once?

PR #2018, opened 2021-04-09
added reset session

o-bahriy:
https://www.pivotaltracker.com/story/show/177629023


aquabu:
@o-bahriy merge conflicts due to some changes in acceptance. Please resolve for the merge. Thanks!

o-bahriy:
@aquabu pulled acceptance and merged

Alexey1100 (commented):
@o-bahriy `activerecord-session_store` docs mention the following configuration on init:

```ruby
Rails.application.config.session_store :active_record_store, :key => '_my_app_session'
```

Looks like we don't have that.
Does it work by default?

o-bahriy:
@Alexey1100 my bad. yes, you are right. added this file

Riddlerrr:
Also, according to the gem's doc we should setup `db:sessions:trim` rake task:

> To avoid your sessions table expanding without limit as it will store expired and potentially sensitive session data, it is strongly recommended in production environments to schedule the db:sessions:trim rake task to run daily. Running bin/rake db:sessions:trim will delete all sessions that have not been updated in the last 30 days. The 30 days cutoff can be changed using the SESSION_DAYS_TRIM_THRESHOLD environment variable.

Riddlerrr:
But I think it should be set up manually using the Heroku scheduler.
@Alexey1100 Can we add the cron job from code somehow?

Alexey1100:
@Riddlerrr nope, the tasks should be defined manually for each env.
We can do it over Heroku API, but it's a bit complicated with our setup.

Alexey1100 (approved):

Alexey1100:
@o-bahriy could you pls merge acceptance to resolve CI issues

o-bahriy:
@Alexey1100 merged

aquabu (approved):

aquabu:
@o-bahriy it changed here: https://github.com/CoMakery/comakery-server/commit/554d236b9fcf5bd2db62e328525ca57e64992ef2#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3

@Riddlerrr can comment on the change. If it needs modification we can address on a follow on commit.

I would be more concerned about a decrease in decimal precision but worth verifying.

Alexey1100:
Yep, all good.
Precision 40 wasn't enough to hold the value, so we updated that a while ago on that branch.

PR #2026, opened 2021-04-10
Add option to disable SSL for Redis

Alexey1100:
Resolves:
https://www.pivotaltracker.com/story/show/177709857

> From version 6 and above, Redis requires using TLS to connect.
However, Heroku does not use SSL internally.

https://help.heroku.com/HC0F8CUS/redis-connection-issues
https://ogirginc.github.io/en/heroku-redis-ssl-error

The issue is only present on Premium Redis add-ons, on free Redis add-ons Heroku does use SSL  🤷
Add ENV `REDIS_SSL_NONE` to trigger the logic.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -63,7 +63,18 @@ module Comakery
     # Use Redis for Cache Store
     redis_provider = ENV.fetch("REDIS_PROVIDER") { "REDIS_URL" }
     redis_url = ENV.fetch(redis_provider) { "redis://localhost:6379/1" }
-    config.cache_store = :redis_cache_store, { url: redis_url, expires_in: 1.hour }
+
+    config.custom_redis_params = {}
+
+    # Disable SSL for Redis in Heroku env:
+    # https://help.heroku.com/HC0F8CUS/redis-connection-issues
+    if ENV["REDIS_SSL_NONE"]
+      config.custom_redis_params = {
+        ssl_params: { verify_mode: OpenSSL::SSL::VERIFY_NONE }
+      }
+    end
+
+    config.cache_store = :redis_cache_store, { url: redis_url, expires_in: 1.hour }.merge(config.custom_redis_params)
     config.action_controller.perform_caching = true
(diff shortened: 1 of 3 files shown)


aquabu (approved):

config/application.rb
```diff
@@ -63,7 +63,18 @@ def self.camelize(string)
...
     redis_url = ENV.fetch(redis_provider) { "redis://localhost:6379/1" }
-    config.cache_store = :redis_cache_store, { url: redis_url, expires_in: 1.hour }
+
+    config.custom_redis_params = {}
+
+    # Disable SSL for Redis in Heroku env:
+    # https://help.heroku.com/HC0F8CUS/redis-connection-issues
+    if ENV["REDIS_SSL_NONE"]
+      config.custom_redis_params = {
+        ssl_params: { verify_mode: OpenSSL::SSL::VERIFY_NONE }
```
aquabu, line 73: @Alexey1100 

To connect directly to the Redis Stunnel via Ruby it appears we want to connect to port `6380` and use the Stunnel scheme as described here: https://devcenter.heroku.com/articles/securing-heroku-redis#connecting-directly-to-stunnel

That Heroku article specifies:
```ruby
require 'uri'
url = URI.parse(ENV["REDIS_URL"])
url.scheme = "rediss"
url.port = Integer(url.port) + 1
$redis = Redis.new(url: url, driver: :ruby, ssl_params: { verify_mode: OpenSSL::SSL::VERIFY_NONE })
```

It seems like this PR's implementation doesn't include:

```ruby
url.scheme = "rediss"
url.port = Integer(url.port) + 1 #which sets the port to 6380
```

Should the `config.custom_redis_params` include those, are they configured somewhere else or not included for a reason?

Thanks!

Alexey1100: Stunnel only works with Redis versions below 6, as noted here:
https://devcenter.heroku.com/articles/securing-heroku-redis

PR #2030, opened 2021-04-12
Redirect to the whitelabel domain if it does not match

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/177715579

I've changed the logic a little bit. I decided to not introduce the new ENV because we already have all necessary data.
We can check `ENV['WHITELABEL']` and if it set and @whitelabel_mission is not set => we will redirect to ENV['APP_HOST']
which is the default domain for Whitelabel mission.

### DIFF
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -228,17 +228,9 @@ describe ApplicationController do
     let(:app_host_var) { 'wl.mission.test' }
     let!(:wl_mission) { create(:whitelabel_mission, whitelabel_domain: app_host_var) }
 
-    around(:each) do |example|
-      prev_wl_env = ENV['WHITELABEL']
-      prev_app_host_env = ENV['APP_HOST']
-
-      ENV['WHITELABEL'] = wl_var
-      ENV['APP_HOST'] = app_host_var
-
-      example.run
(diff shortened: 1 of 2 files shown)


spec/controllers/application_controller_spec.rb
```diff
@@ -222,4 +222,39 @@ def show
...
+  describe '#redirect_to_default_whitelabel_domain' do
+    let(:wl_var) { 'true' }
+    let(:app_host_var) { 'wl.mission.test' }
+    let!(:wl_mission) { create(:whitelabel_mission, whitelabel_domain: app_host_var) }
+
+    around(:each) do |example|
+      prev_wl_env = ENV['WHITELABEL']
+      prev_app_host_env = ENV['APP_HOST']
+
+      ENV['WHITELABEL'] = wl_var
```
Alexey1100, line 235: I think it's better to stub the ENV, to avoid issues when tests are running in parallel.

```ruby
stub_const('ENV', ENV.to_hash.merge('WHITELABEL' => wl_var))
```

Riddlerrr: @Alexey1100 Nice, thank you for the tip 👍

aquabu (approved):

PR #2045, opened 2021-04-16
Alias Account Token Records as Wallet Transfer Rules

olehyev138:
Alias the account_token_records API endpoint to wallet_transfer_rules and update the docs

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -189,6 +189,11 @@ Rails.application.routes.draw do
             delete :index, action: :destroy_all
           end
         end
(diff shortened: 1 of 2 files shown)


aquabu:
@olehyev138 does the previous path still work or does this replace it entirely? Some users are depending on the old path so this should be an alias. Just want to double check this won't break things for them.

olehyev138:
@aquabu Yes, the previous path still works, this is just an alias that leads to the same controller, I have checked this.

aquabu (approved):

PR #2053, opened 2021-04-21
Accounts page

zhinfuhnya:
[#177788836](https://www.pivotaltracker.com/n/projects/2167210/stories/177788836)


app/models/wallet.rb
```diff
@@ -25,6 +29,10 @@ class Wallet < ApplicationRecord
   before_create :set_primary_flag
   after_commit :mark_first_wallet_as_primary, on: [:destroy], if: :primary_wallet?
 
+  after_create_commit { broadcast_append_later_to 'wallets' if account_whitelabel? }
```
Alexey1100, line 32: Looks like we need to include whitelabel mission id into turbo stream id.
Otherwise accounts would be leaking across all whitelabel missions.

app/views/wallets/_wallet.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= content_tag :tr, id: dom_id(wallet) do %>
+  <td class="h12 font-weight-medium n10-color"><%= "#{wallet.account.first_name} #{wallet.account.last_name}" %></td>
+  <td class="h12 font-weight-medium n10-color"><%= wallet.account.email %></td>
+  <td class="h12 font-weight-medium n10-color"><%= wallet._blockchain %></td>
```
Alexey1100, line 4: Should be `wallet.blockchain.name` for the UI.

app/controllers/accounts_controller.rb
```diff
@@ -6,12 +6,22 @@ class AccountsController < ApplicationController
...
+  skip_after_action :verify_authorized, :verify_policy_scoped, only: %i[index new create confirm confirm_authentication show download_data]
 
   before_action :redirect_if_signed_in, only: %i[new create]
 
   layout 'legacy'
 
+  def index
+    authorize(current_user, :index?)
+
+    @wallets = Wallet
```
Alexey1100, line 18: Should only include wallets from `@whitelabel_mission.managed_accounts`

Alexey1100 (changes requested):
Needs a couple of changes and some specs:
- Acceptance spec which actually renders the page under different conditions (Similar to `spec/features/dashboard/accounts/index_spec.rb`)
- Policy spec coverage

aquabu (approved):

app/controllers/accounts_controller.rb
```diff
@@ -15,11 +16,13 @@ class AccountsController < ApplicationController
...
 
-    @wallets = Wallet
-               .with_whitelabel_account
-               .includes(account: :latest_verification)
-               .page(params[:page])
-               .per(10)
+    @wl_account_wallets = @whitelabel_mission
+                          .managed_accounts
+                          .includes(:latest_verification, :wallets)
+                          .map(&:wallets)
```
Alexey1100, line 22: The `.map` will load the complete query result in memory as an array, which might be more than 10k records :)
We need to do wallet load on database side

app/models/account.rb
```diff
@@ -377,4 +387,22 @@ def populate_awards
         a.update(account: self, email: nil)
       end
     end
+
+    def broadcast_update_wl_account_wallet
+      wallets.each do |wallet|
+        broadcast_replace_to 'wl_account_wallets',
```
Alexey1100, line 393: We need to include `managed_mission.id` in the web socket channel name `wl_account_wallets`, not in the DOM target.
Otherwise everyone who connected to `wl_account_wallets` channel will see updates for all of the accounts for every white label instance.

app/models/verification.rb
```diff
@@ -15,6 +17,15 @@ def failed?
 
   private
 
+    def broadcast_update_wl_account_wallet
+      account.wallets.each do |wallet|
+        broadcast_replace_to 'wl_account_wallets',
```
Alexey1100, line 22: Same here:

> We need to include managed_mission.id in the web socket channel name wl_account_wallets, not in the DOM target.

app/models/wallet.rb
```diff
@@ -110,7 +106,24 @@ def empty_address_allowed?
       ore_id? && (wallet_provisions.empty? || wallet_provisions.any?(&:pending?))
     end
 
-    def account_whitelabel?
-      account&.managed_mission&.whitelabel?
+    def broadcast_create_wl_account_wallet
+      broadcast_append_later_to 'wl_account_wallets',
```
Alexey1100, line 110: Same here:

> We need to include managed_mission.id in the web socket channel name wl_account_wallets, not in the DOM target.

app/models/wallet.rb
```diff
@@ -110,7 +106,24 @@ def empty_address_allowed?
...
+                                target: "wl_#{account.managed_mission.id}_account_#{account.id}_wallet_#{id}",
+                                partial: 'accounts/partials/index/wl_account_wallet',
+                                locals: { wl_account_wallet: self }
+    end
+
+    def broadcast_update_wl_account_wallet
+      broadcast_replace_to 'wl_account_wallets',
```
Alexey1100, line 117: Same here:

> We need to include managed_mission.id in the web socket channel name wl_account_wallets, not in the DOM target.

Alexey1100 (changes requested):

aquabu:
@Alexey1100 can you check if your issue is resolved on this PR?

Alexey1100 (approved):

PR #2057, opened 2021-04-22
Admin access management

zhinfuhnya:


app/controllers/dashboard/accounts_controller.rb
```diff
@@ -89,21 +89,12 @@ def set_accounts
...
-
-      if @project.token&.token_type&.operates_with_account_records?
-        @q = @q.joins('inner join account_token_records on account_token_records.token_id = projects.token_id and account_token_records.account_id = accounts.id')
-               .includes(
-                 account_token_records: [:reg_group]
-               )
-      end
+      @q = Account
+           .joins(:project_roles)
+           .where(project_roles: { project: @project.object })
```
Riddlerrr, line 94: What the `@project.object` is mean? I can't found that method.

Riddlerrr (changes requested):
Small visual bug. It shows as current "accounts" and "my projects" in the same time:
CoMakery - Achieve Big Missions 2021-05-03 11-03-39

On the account's page it shows useless pencil. When I click it nothing happens:
CoMakery - Achieve Big Missions 2021-05-03 11-20-31

This button also doesn't work:
CoMakery Project 2021-05-03 11-26-38

Also, something happened with the header (comakery logo) and the footer:
CoMakery - Achieve Big Missions 2021-05-03 11-29-21

Overall the code looks really well, great job 👍

aquabu:
I think the two account links correspond to the organizational Accounts and the Project Accounts.

app/controllers/dashboard/accounts_controller.rb
zhinfuhnya: fetch project record from decorator
https://github.com/CoMakery/comakery-server/blob/acceptance/app/controllers/application_controller.rb#L142

Riddlerrr: oh, ok, it's decorator

zhinfuhnya: regarding to pencil on accounts page: our changes do not affect this functionality, most likely requires a separate story.

zhinfuhnya: I would like to clarify: are you trying to unfollow from the project as an admin? (because there is a validation for unfollow, in which there is a restriction on the destroy of interests for admins)

Riddlerrr: @zhinfuhnya Please ignore the "unfollow" button issue. It works the same as on acceptance branch, so it's not related this changed and should be fixed seeparately.

Riddlerrr: I've removed this screenshot from my initial comment to not confuse anyone.

Riddlerrr (approved):

aquabu (changes requested):
@zhinfuhnya please fix the accounts edit issue on the accounts page for the CoMakery Security Token token types.

app/assets/stylesheets/upside.scss
```diff
@@ -10794,14 +10800,14 @@ Card tabs
...
 
-.icon {
-  width: 1.25rem;
-  height: 1.25rem;
-  font-size: 1.25rem;
-  vertical-align: bottom;
-  stroke-width: 1.5; }
-  .icon:hover {
-    text-decoration: none; }
+//.icon {
```
Alexey1100, line 10803: Looks like it should be removed

app/controllers/projects/invites_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+      else
+        format.json { render json: { errors: send_invite_result.errors }, status: :unprocessable_entity }
+      end
+    end
+  end
+
+  private
+
+    def find_project
+      @project = Project.find(params[:project_id])
```
Alexey1100, line 23: `@project_scope.find` see comment above.

app/interactors/send_invite.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    invite = context.project.project_roles.new(account: account, role: context.params[:role])
+
+    context.fail!(errors: invite.errors.full_messages) unless invite.save
+  end
+
+  private
+
+    def account
+      Account.find_by(email: context.params[:email])
```
Alexey1100, line 15: Emails are not unique across accounts table, so we should also supply whitelabel mission.
Pls see existing implementation of Admins page.

app/javascript/controllers/permissions/form_controller.js
```diff
@@ -0,0 +1,51 @@
...
+
+    Rails.ajax({
+      type: this.element.attributes.method.value,
+      url: this.element.attributes.action.value,
+      data: new FormData(this.element),
+      success: (response) => {
+        if (response.message) {
+          this._addFlashMessage('notice', response.message)
+        } else {
+          this._addFlashMessage('error', 'You are not authorized to perform this action')
```
Alexey1100, line 26: I'm thinking about how we can have backend as a single source of truth for all error messages 🤔
Maybe backend can render a turbo frame which will trigger a custom stimulus controller to publish the message.

What do you think?

app/models/project_role.rb
```diff
@@ -0,0 +1,10 @@
+class ProjectRole < ApplicationRecord
+  belongs_to :account
+  belongs_to :project
+
+  # rubocop:todo Rails/UniqueValidationWithoutIndex
+  validates :project_id, uniqueness: { scope: %i[account_id], message: 'already has a role in project' }
```
Alexey1100, line 6: We need to add the index here.
See wallets table for an example of scoped unique index.

spec/controllers/projects/invites_controller_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
+
+RSpec.describe Projects::InvitesController, type: :controller do
+  let!(:admin) { create(:account, comakery_admin: true) }
```
Alexey1100, line 4: Just to avoid confusion :) `comakery_admin` is not related to projects.
It's a flag to access Missions admin and Tokens admin.

spec/models/project_role_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe ProjectRole, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
Alexey1100, line 4: Some shoulda matchers covering validations, enum and associations would be nice here.

Alexey1100 (changes requested):
Good job :)

I think there are couple improvements needed:

app/javascript/controllers/accept_turbo_stream_controller.js
```diff
@@ -0,0 +1,29 @@
...
+      'turbo:before-fetch-request',
+      this.addTurboStreamHeaders
+    )
+  }
+
+  addTurboStreamHeaders(event) {
+    const { headers } = event.detail.fetchOptions || {}
+
+    if (headers) {
+      headers.Accept = ['text/vnd.turbo-stream.html', headers.Accept].join(', ')
```
Alexey1100, line 26: Interesting..
Why it doesn't happen by default.

db/schema.rb
```diff
@@ -334,9 +337,6 @@
     t.index ["project_id"], name: "index_channels_on_project_id"
     t.index ["team_id"], name: "index_channels_on_team_id"
   end
-  
-  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
```
Alexey1100, line 338: Hmm, this is not the first time we are seeing such weird schema updates.
Wondering what's the reason for that.

Alexey1100 (changes requested):
Pls fix schema.rb

aquabu (approved):

app/models/project_role.rb
```diff
@@ -0,0 +1,35 @@
...
+    def broadcast_update
+      broadcast_replace_to "project_#{project.id}_accounts",
+                           target: "project_#{project.id}_account_#{account.id}",
+                           partial: 'dashboard/accounts/account',
+                           locals: { project_role: self }
+    end
+
+    def broadcast_create
+      broadcast_append_to "project_#{project.id}_accounts",
+                          target: "project_#{project.id}_accounts",
```
Alexey1100, line 24: I think there's a little issue which I don't know yet how to handle.
It will append the records to the target table regardless current filtering/sorting/pagination.

So if you are on a page it will append the new records to the end of the page (wrongly), but if you refresh the page records will be gone, depending on what filter/sorting/page you have. 

Maybe we need to disable that until we figure the pattern we need to use in such cases.

app/models/project.rb
```diff
@@ -21,6 +21,10 @@ class Project < ApplicationRecord
   belongs_to :token, optional: true, touch: true
   has_many :interests # rubocop:todo Rails/HasManyOrHasOneDependent
   has_many :interested, -> { distinct }, through: :interests, source: :account
+  has_many :project_roles # rubocop:todo Rails/HasManyOrHasOneDependent
```
Alexey1100, line 24: Non blocking.
We need `dependent: nullify` here.

app/models/project_role.rb
zhinfuhnya: Thanks for the note! :)

This issue will be solved if we use `broadcast_append_later_to` (the record will be attached at the end of the table when we switch to the next page)

Alexey1100 (approved):

app/models/project_role.rb
Alexey1100: Hmm I think the difference between `broadcast_append_to` and `broadcast_append_later_to` is that the second schedules a job for the broadcast while the first one does it inline.
So the result will be the same from user perspective.

Alexey1100:
@zhinfuhnya good
pls make sure to update `Api::V1::AccountsController#create` as well, since we have some logic around interests there as well.

app/models/project_role.rb
Alexey1100: Maybe we can update frame id based on current filter/page, so the logic will be only triggered on a first page with default sorting and without filters.

Otherwise it will be super complex to handle :)

zhinfuhnya:
@Alexey1100 @aquabu I think it would be better to wait for the PR with migrations of interests before merging this branch into the acceptance.

PR #2059, opened 2021-04-22
updated transfer columns

olehyev138:
Update project transfer page table columns

See attached image for how it will look. It's not pixel perfect but is pretty close to how it will look.

Acceptance:

- Rename "Transfer Name" to "Transfer"

- Rename "Account" column to "To"

- Include the blockchain address of the recipient in the "To" column. This is the users primary wallet address for the project token's blockchain network

- Rename "Transferred By" to "From". The address is a link to the blockchain explorer.

- Only display the "Total Amount" of the transfer - not quantity or amount

- Move the verification to it's own column

- Status no longer includes the transaction "tx" link

- All status wallet payment and cancelation steps still work correctly

- History column includes the "tx" transaction link

- Transferred remains

- Remove created date


app/views/dashboard/transfers/_transfer.html.erb
```diff
@@ -18,7 +18,13 @@
     <label>account</label>
 
     <% if transfer.account %>
-      <%= render partial: "account", locals: {account: transfer.account} %>
+      <% sender_wallet_address = nil %>
```
aquabu, line 21: This is too much logic for the view. I recommend moving the logic to the decorator classes, unit testing the decorator classes and avoiding setting any variables in the view.

app/views/dashboard/transfers/_transfer.html.erb
```diff
@@ -38,48 +44,45 @@
...
+      <% account = transfer.account.decorate %>
+      <%= image_pack_tag "media/src/images/styleguide/icons/verification-#{account.verification_state}.svg", size: "15x15" %>
+      <%= account.verification_state&.capitalize %>
+    </div>
+  </div>
+
+  <div class="transfers-table__transfer__status">
+    <label>status</label>
+
+    <% if @project.token %>
```
aquabu, line 66: Just noticing that this code has too much logic for the view - much of which was already there. The view should make simpler checks and should avoid this amount of nested logic too avoid bugs.

I recommend moving the logic to decorators and adding additional unit testing it there.

aquabu (changes requested):

app/views/dashboard/transfers/_transfer.html.erb
aquabu, line 21: should not set variables in the view

app/views/dashboard/transfers/_transfer.html.erb
```diff
@@ -18,7 +18,13 @@
     <label>account</label>
 
     <% if transfer.account %>
-      <%= render partial: "account", locals: {account: transfer.account} %>
+      <% sender_wallet_address = nil %>
+      <% sender_wallet_url = nil %>
+      <% if transfer.paid? %>
+        <% sender_wallet_address = transfer.blockchain_transactions.succeed.last&.source %>
```
aquabu, line 24: move this `sender_wallet_address` logic to `transfer.sender_wallet_address` transfer decorator function with a unit test

app/views/dashboard/transfers/_transfer.html.erb
```diff
@@ -18,7 +18,13 @@
...
+        <% sender_wallet_url = sender_wallet_address && transfer.token ? transfer.token.blockchain.url_for_address_human(sender_wallet_address) : nil %>
```
aquabu, line 25: move this `sender_wallet_url` logic to `transfer.sender_wallet_url` transfer decorator function with a unit test

app/views/dashboard/transfers/_transfer.html.erb
```diff
@@ -18,7 +18,13 @@
...
+      <% end %>
+      <%= render partial: "account", locals: {account: transfer.account, wallet: sender_wallet_address, wallet_url: sender_wallet_url} %>
```
aquabu, line 27: use the `transfer` decorator methods for `transfer.sender_wallet_address` and `transfer.sender_wallet_url`

app/views/dashboard/transfers/_transfer.html.erb
aquabu, line 21: confirm that there is an feature spec that tests this larger if statement

app/views/dashboard/transfers/_transfer.html.erb
```diff
@@ -38,48 +44,45 @@
...
+      <%= account.verification_state&.capitalize %>
+    </div>
+  </div>
+
+  <div class="transfers-table__transfer__status">
+    <label>status</label>
+
+    <% if @project.token %>
+    <% if transfer.project.token&._token_type? %>
+    <% if current_account&.decorate&.can_send_awards?(transfer.project) %>
```
aquabu, line 68: `current_account&.decorate` can happen in the controller 
`can_send_awards?` should be a AwardPolicy check, also confirm it is unit tested.

app/views/dashboard/transfers/_transfer.html.erb
```diff
@@ -38,48 +44,45 @@
...
+      <% if transfer.token&.blockchain&.supported_by_ore_id? && current_account&.ore_id_account.nil? %>
```
aquabu, line 69: this can be moved to a `transfer` decorator method `transfer.payable_by_ore_id?(current_account)` and unit tested

app/views/dashboard/transfers/_transfer.html.erb
aquabu, line 68: In addition to unit testing make sure each of these conditions is testing in an feature integration test

app/views/dashboard/transfers/_transfer.html.erb
```diff
@@ -38,48 +44,45 @@
...
+        <div class="transfer-button">
+          <%= link_to 'link OREID', wallets_path %>
+        </div>
+      <% else %>
+        <%= render partial: "shared/transfer_button_admin", locals: { transfer: transfer } %>
       <% end %>
+    <% else %>
+      <% if transfer.recipient_address.blank? && current_account == transfer.account %>
```
aquabu, line 77: This logic looks overly complex and could be a source of bugs. Simplify this by removing this `set wallet` if branch and render `shared/transfer_button_public` which will display "needs wallet" if it needs one.

app/views/dashboard/transfers/_transfer.html.erb
```diff
@@ -92,14 +95,26 @@
...
+      –
+    <% end %>
+  </div>
 
-      <%= render partial: "transfer_history_button", locals: { transfer: transfer } %>
+  <div class="transfers-table__transfer__history">
+    <label>history</label>
+
+    <% if @project.token %>
+      <% if transfer.blockchain_transactions.size > 0 %>
```
aquabu, line 107: refactor `transfer.blockchain_transactions.size > 0` to the transfer decorator as `transfer.has_blockchain_transactions?` and unit test it.

aquabu (changes requested):

app/views/dashboard/transfers/_transfer.html.erb
olehyev138: so just remove the whole if-else statement and just put in transfer_button_public render?

olehyev138:
to solve some Test failed, need @Alexey1100's help

app/views/dashboard/transfers/_transfer.html.erb
aquabu: yes I think so. @Alexey1100 does that sound correct?

Alexey1100: It's a bit tricky yes, since `shared/transfer_button_public` is rendered for broadcast and doesn't know about current user and it's wallet. That's why the set wallet was extracted from there.

@aquabu You are right, the `needs wallet` message still will be displayed, but we will not be able to link contributor to wallets page as currently.

aquabu: ok - I think we can resolve in a future PR.

PR #2064, opened 2021-04-23
remove whitelabel-projects-task

olehyev138:
When the WHITELABEL environment variable is set to true then the project "Tasks" menu item should not appear

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -232,6 +232,7 @@ class ProjectsController < ApplicationController
 
     def set_show_props # rubocop:todo Metrics/CyclomaticComplexity
       @props = {
+        whitelabel: ENV['WHITELABEL'] || false,
         tasks_by_specialty: @project.ready_tasks_by_specialty.map do |specialty, awards|
           [
             specialty&.name&.downcase || 'general',
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -229,21 +229,25 @@ class ProjectsController < ApplicationController
     def project_header
       @project ? @project.decorate.header_props : { image_url: helpers.image_url('default_project.jpg') }
     end
+    
+    def project_tasks_by_specialty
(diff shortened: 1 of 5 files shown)


aquabu (changes requested):
@olehyev138 
* please add an rspec integration test for when the menu item appears and when it does not appear
* also the URL should not be accessible, this should also have a test

aquabu:
@olehyev138 this still needs the rspec tests mentioned above. Thanks.

aquabu (approved):

app/controllers/projects_controller.rb
```diff
@@ -230,19 +230,24 @@ def project_header
...
-            specialty&.name&.downcase || 'general',
-            awards.map do |task|
-              task_to_props(task).merge(
-                allowed_to_start: policy(task).start?,
-                reached_maximum_assignments: task.reached_maximum_assignments_for?(current_account)
-              )
-            end
-          ]
-        end,
+        whitelabel: ENV['WHITELABEL'] || false,
```
Alexey1100, line 249: We should check `@whitelabel_mission` variable here which indicates that request comes from whitelabel.

`ENV['WHITELABEL']` is only used for a specific case setup when the complete deployment is whitelabel.

app/policies/project_policy.rb
```diff
@@ -48,7 +48,9 @@ def show_contributions?
   end
 
   def show_award_types?
-    show? || unlisted?
+    return false if ENV['WHITELABEL'] == true
```
Alexey1100, line 51: Same here, we need to pass @whitelabel_mission into policy.

spec/features/tasks/project_tasks.spec.rb
```diff
@@ -7,6 +7,24 @@
     ready_task.project.update(visibility: :public_listed)
   end
 
+  it 'shows tasks link if env whitelabel is not true' do
+    login(ready_task.account)
```
Alexey1100, line 11: Hmm I think these specs might not be working correctly...
Since we are not switching env to whitelabel between examples.

PR #2083, opened 2021-05-07
Optimize token balances query

zhinfuhnya:


db/schema.rb
```diff
@@ -438,6 +439,18 @@
     t.index ["project_id"], name: "index_payments_on_project_id"
   end
 
+  create_table "project_roles", force: :cascade do |t|
```
Riddlerrr, line 442: @kolishchak This is not an actual change for this branch. Please update the schema file to add just one index.

Riddlerrr (changes requested):

zhinfuhnya:
According to my benchmarks this will speed up query execution by half.

Riddlerrr (approved):

PR #2095, opened 2021-05-11
Migrate interests to project roles

zhinfuhnya:


app/models/project.rb
```diff
@@ -103,7 +105,7 @@ def self.assign_project_owner_from(project_or_project_id, email)
     project.safe_add_admin(previous_owner)
     project.account_id = new_owner.id
     project.admins.delete(new_owner)
-    project.safe_add_interested(new_owner)
+    project.add_account(new_owner)
```
o-bahriy, line 108: here it should be add_admin

db/data_migrations/20210512175535_update_project_roles_counter.rb
```diff
@@ -0,0 +1,10 @@
+class UpdateProjectRolesCounter < ActiveRecord::DataMigration
+  def up
```
aquabu, line 2: The failure rate for data migrations is pretty high because there's often unexpected data values or nils. Please download the demo server data and run the data migrations. The also write specs to verify it works for the expected case and any failures you find with the demo data. Thanks!

app/controllers/accounts_controller.rb
```diff
@@ -63,7 +63,7 @@ def create
       UserMailer.with(whitelabel_mission: @whitelabel_mission).confirm_email(@account).deliver
 
       Project.where(auto_add_interest: true).each do |auto_interest_project|
-        auto_interest_project.safe_add_interested(@account)
+        auto_interest_project.add_account(@account)
```
aquabu, line 66: If there isn't a spec it would good to write one verifying that it's not possible to get added more than once. Historically there have been some bugs around displaying the same account more than once on the accounts page.

aquabu (changes requested):

db/data_migrations/20210512175535_update_project_roles_counter.rb
zhinfuhnya: Migration was run on actual data 👌 

Also added spec https://github.com/CoMakery/comakery-server/pull/2095/commits/2bf0b09dbb4a86d1f9f21fb881f684a0fc031ffa

app/controllers/accounts_controller.rb
zhinfuhnya: We already have specifications that say we cannot add more than one account.

aquabu (approved):

o-bahriy (approved):

zhinfuhnya:
Includes the following PR's:
https://github.com/CoMakery/comakery-server/pull/2097
https://github.com/CoMakery/comakery-server/pull/2120

app/controllers/projects_controller.rb
```diff
@@ -19,18 +19,30 @@ class ProjectsController < ApplicationController
...
     end
 
     @my_project_contributors = TopContributors.call(projects: @my_projects).contributors
     @team_project_contributors = TopContributors.call(projects: @team_projects).contributors
-    @interested_project_contributors = TopContributors.call(projects: @interested_projects).contributors
+    @involved_project_contributors = TopContributors.call(projects: @involved_projects).contributors
     @archived_project_contributors = TopContributors.call(projects: @archived_projects).contributors
   end
 
+  def awards
```
Alexey1100, line 34: This page was deleted recently, so it shouldn't be here.

spec/controllers/dashboard/accesses_controller_spec.rb
```diff
@@ -10,124 +10,21 @@
...
-      post :add_admin, params: { project_id: project.to_param, email: project_admin.email }
-      project.reload
-
-      expect(response).to redirect_to(project_dashboard_accesses_path(project))
-      expect(flash[:notice]).not_to be_nil
-      expect(project.admins).to match_array([project_admin])
-      expect(project.interested).to include(project_admin)
-    end
-
-    context 'with whitelabel mission' do
```
Alexey1100, line 36: Hmm do we still have that coverage somewhere else?
I can't find it :)

Looks like the interaction is moved into SendInvite class but it's specs don't cover whitelabel interactions.

Alexey1100 (changes requested):

Alexey1100 (approved):

aquabu (approved):

PR #2105, opened 2021-05-12
Reorganize/transfer

olehyev138:
Acceptance:

User search works
Transfers date sets the calendar dates
Form elements are in the correct locations
Can filter by status and it updates the charts and the transactions table
Can filter by category and this updates the chart and the transactions table
Category totals and percentages are correct
Percentages add up to 100%
Create new transfer includes the current transfer types


aquabu:
@olehyev138 looks like there are some failing specs on this one that need to be resolved.

PR #2112, opened 2021-05-14
Cache token balances

zhinfuhnya:


app/views/api/v1/accounts/token_balances.json.jbuilder
```diff
@@ -1,17 +1,19 @@
 json.array! @account.balances.includes(:wallet, :token).map do |balance|
-  json.token do
-    json.partial! 'api/v1/tokens/token', token: balance.token
-  end
+  json.cache! balance do
+    json.token do
+      json.partial! 'api/v1/tokens/token', token: balance.token
```
zhinfuhnya, line 4: Should we remove token in this PR?

PR #2113, opened 2021-05-14
Update project role policy

zhinfuhnya:


app/policies/project_role_policy.rb
```diff
@@ -5,7 +5,16 @@ def initialize(account, project_role)
   end
 
   def update?
-    account.present? && project_admin? && !own_project_role?
+    return false if account.nil? || project_role.account == account
+
+    case project_role.role.to_sym
+    when :admin
```
aquabu, line 11: Something about this looks odd. If you are `:admin` and `project_admin?` but not `project_owner?` should you be able to update? 

Maybe I'm missing something but it seems that with this logic an `:admin` with `project_admin?` would not be able to update but `:interested` `project_admin?` would.

aquabu (changes requested):

app/policies/project_role_policy.rb
zhinfuhnya: Accounts with the role of admin can only edit the superior (project owner). That is, admins can not change each other's role (which makes sense to me).

Only owners and admins can update roles of interested and observers. 

Do you agree with me?

zhinfuhnya: > If you are :admin and project_admin? but not project_owner? should you be able to update?

it is impossible because the condition to check the editing of their role is met, admins can't edit themselves.

Alexey1100: @zhinfuhnya yes, you're right.

I think the confusion here is because the method is a bit hard to read.
Maybe instead of the case condition we should add a bunch of helpers with clear naming.
E.g:

`can_edit_admin_role?`
`can_edit_interested_role?`
`can_edit_observer_role?`
`can_edit_own_role?`

zhinfuhnya: @Alexey1100 agree with you.
Added review fixes.

aquabu (approved):

PR #2116, opened 2021-05-15
Admin can download a CSV of all transfers from the transfers page

shoaibmalik786:


app/decorators/project_decorator.rb
```diff
@@ -273,6 +273,60 @@ def ratio_pretty(value, total)
     end
   end
 
+  # rubocop:todo Metrics/PerceivedComplexity
+  def project_transfers_csv_data # rubocop:todo Metrics/CyclomaticComplexity
```
zhinfuhnya, line 277: It would be good to move the transfers export logic to services, e.g. `TransfersExporter`
And also add specs :)

zhinfuhnya (changes requested):

app/decorators/project_decorator.rb
shoaibmalik786: Will do.

Alexey1100 (changes requested):
Needs some updates as mentioned below.

Also this link looks like a great possibility for DDoS, especially on project with large amount of transfers.
We should at least cache/store it and not allow to re-run export too often (e.g once per hour).

Ideally we would run the export in a background job and add the file link to the page via turbo stream when the export is finished and stored in active storage.
Another option is sending file to admin email address, when the export is finished.

shoaibmalik786:
> Needs some updates as mentioned below.
> 
> Also this link looks like a great possibility for DDoS, especially on project with large amount of transfers.
> We should at least cache/store it and not allow to re-run export too often (e.g once per hour).
> 
> Ideally we would run the export in a background job and add the file link to the page via turbo stream when the export is finished and stored in active storage.
> Another option is sending file to admin email address, when the export is finished.



> Needs some updates as mentioned below.
> 
> Also this link looks like a great possibility for DDoS, especially on project with large amount of transfers.
> We should at least cache/store it and not allow to re-run export too often (e.g once per hour).
> 
> Ideally we would run the export in a background job and add the file link to the page via turbo stream when the export is finished and stored in active storage.
> Another option is sending file to admin email address, when the export is finished.

@Alexey1100 You mean we will save this CSV in the db ? if yes , Should i create the new modal for this? Please suggest.

Alexey1100:
@shoaibmalik786 I think we can use active storage for that.

Basically we can add `has_one_attached :transfers_csv` to `Project` and a method, something like `update_transfers_report`. This method would run the export and update the file or skip the run if the report was already updated less than an hour ago.

On the transfers page we would have a link to the file as well as update button, triggering a background job which runs `update_transfers_report`

PR #2119, opened 2021-05-17
Sync locked and unlocked balance

Alexey1100:
Resolves:
https://www.pivotaltracker.com/story/show/177968991
https://www.pivotaltracker.com/story/show/177986087

- Add balance helpers to `TokenType::TokenReleaseSchedule`
- Add balance helpers to `Balance`
- Return `lockup_schedule_ids` only for paid transfers


zhinfuhnya (approved):

app/models/balance.rb
```diff
@@ -25,15 +25,35 @@ def unlocked_value
...
+  def blockchain_balance_base_unit_locked_value
+    @blockchain_balance_base_unit_locked_value ||= if token._token_type_token_release_schedule?
+      token.blockchain_locked_balance(wallet.address)
+    else
+      0
+    end
+  end
+
+  def blockchain_balance_base_unit_unlocked_value
+    @blockchain_balance_base_unit_unlocked_value ||= if token._token_type_token_release_schedule?
```
aquabu, line 44: Following the pattern of refactoring case statements switching on type to calls to classes with a common interface method, these should be methods on the token model and should appear here as something like:
`token.blockchain_unlocked_balance(wallet.address)`
`token.blockchain_locked_balance(wallet.address)`

With no case statement. There may be a similar case statement type to common interface method refactoring on the token model as well.

app/models/balance.rb
```diff
@@ -25,15 +25,35 @@ def unlocked_value
...
+      token.blockchain_unlocked_balance(wallet.address)
+    else
+      blockchain_balance_base_unit_value
+    end
+  end
+
   def sync_with_blockchain!
-    update(base_unit_value: blockchain_balance_base_unit_value)
+    update!(
+      base_unit_value: blockchain_balance_base_unit_value,
```
aquabu, line 53: With the refactoring above, methods like `blockchain_balance_base_unit_unlocked_value` would be deleted and this would just be something like:
```
base_unit_value: token.balance_base_unit_value,
base_unit_locked_value: token.balance_base_unit_locked_value,
base_unit_unlocked_value: token.balance_base_unit_unlocked_value
```

aquabu (approved):

PR #2120, opened 2021-05-17
Visibility restrictions by project roles 

zhinfuhnya:


app/policies/project_policy.rb
```diff
@@ -57,6 +57,14 @@ def show_transfer_rules?
     show_contributions? && project.supports_transfer_rules?
   end
 
+  def accounts?
+    show_contributions? || project_observer? || project_interested?
```
aquabu, line 61: users who are only `interested` shouldn't be able to see the accounts. They would only be able to see the accounts if `show_contributions?` is true.

app/policies/project_policy.rb
```diff
@@ -57,6 +57,14 @@ def show_transfer_rules?
...
+  end
+
+  def transfers?
+    show_contributions? || project_observer? || project_interested?
```
aquabu, line 65: same as for `#accounts?`

you can delete `project_interested?` here. Interested can only see transfers when `show_contributions?` is true.

app/views/dashboard/accounts/_account.html.erb
```diff
@@ -159,25 +160,28 @@
...
-                        method: :get,
-                        data: { controller: 'accept-turbo-stream' },
-                        id: 'change_permissions_btn',
-                        class: 'dropdown-item' %>
+
+      <% if @project_policy.change_permissions? %>
+        <div class="transfers-table__transfer__settings">
+          <div class="dropdown">
+            <a class="dropdown" data-id="settings" data-boundary="viewport" data-toggle="dropdown">
+              <svg class="n9-color" width="18" height="4" viewBox="0 0 18 4" fill="none" xmlns="http://www.w3.org/2000/svg">
```
aquabu, line 168: Maybe move this svg to its own file or partial?

aquabu (changes requested):

app/views/dashboard/accounts/_account.html.erb
zhinfuhnya: I think we can leave it as it is, since we do not reuse svg anywhere else.

PR #2128, opened 2021-05-19
Optimize token balances

zhinfuhnya:


Riddlerrr (approved):
👍

Alexey1100 (changes requested):
Pls check the caching key with pagination, as mentioned on call and regenerate api docs, so pagination headers would be displayed in the docs.

public/doc/api/v1/ii._accounts/token_balances.html
```diff
@@ -134,6 +134,14 @@ <h3>Parameters</h3>
                   <span class="description">account id</span>
                 </td>
               </tr>
+              <tr>
+                <td>
+                  <span class="name">page</span>
```
Alexey1100, line 139: Hmm where does this update come from?
There's no doc spec changes included in the PR.

Alexey1100 (changes requested):

Alexey1100 (approved):

aquabu (approved):

PR #2130, opened 2021-05-19
Header project token details

zhinfuhnya:


app/javascript/components/layouts/ProjectSetupHeader.jsx
```diff
@@ -215,6 +215,10 @@ class ProjectSetupHeader extends React.Component {
 
         <ProjectInfo>
           <h1>{project && project.title || 'New Project'}</h1>
+          <div style={{ display: 'flex' }}>
```
Alexey1100, line 218: Should only be displayed if project token is present.

Alexey1100 (changes requested):
Will crash on projects without token.

app/decorators/project_decorator.rb
```diff
@@ -93,7 +93,7 @@ def send_coins?
     token&._token_type? && %w[eth btc ada qtum eos xtz].include?(token&._token_type)
   end
 
-  def header_props
+  def header_props # rubocop:todo Metrics/CyclomaticComplexity
```
aquabu, line 96: Please refactor out some of the conditionals instead of disabling Rubocop. This will help us keep the codebase more maintainable long term.

I think in this case the 5 `token&.` statements may just need to be a single branching conditional on something like `if token.present?`

aquabu (changes requested):
refactor cyclomatic complexity

app/decorators/project_decorator.rb
```diff
@@ -122,6 +122,23 @@ def header_props
...
+
+    {
+      token: {
+        name: token.name,
+        symbol: token.symbol,
+        network: token.blockchain_network,
+        address: token.contract_address,
+        logo_url: GetImageVariantPath.call(
+          attachment: token.logo_image,
+          esize_to_fill: [100, 100]
```
Alexey1100, line 139: Typo here

app/decorators/project_decorator.rb
```diff
@@ -122,6 +122,23 @@ def header_props
...
+  end
+
+  def token_props
+    return {} if token.nil?
+
+    {
+      token: {
+        name: token.name,
+        symbol: token.symbol,
+        network: token.blockchain_network,
```
Alexey1100, line 135: This field is deprecated, should use `token.blockchain.name` instead

Alexey1100 (changes requested):

app/decorators/project_decorator.rb
Alexey1100: Pls add a spec so this error would be caught.

aquabu (approved):

Alexey1100 (approved):

PR #2143, opened 2021-05-27
Observer role visibility

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/177789009

Updates includes:
- Added policies for the project header
- Observer has access to transfers if the project is public and doesn't require confidentiality
- Updated policy specs


zhinfuhnya:
@aquabu @Alexey1100 @Riddlerrr I want to confirm my changes regarding adding accounts to the awards, originally we added them as interested but since we restrict access to interested in the transfers I changed logic to adding an account as observer.

app/policies/project_policy.rb
```diff
@@ -62,7 +62,9 @@ def accounts?
   end
 
   def transfers?
-    show_contributions? || project_observer?
+    return false if project_interested?
+
+    edit? || (project_observer? ? show_contributions_to_everyone? : show_contributions?)
```
Riddlerrr, line 67: I think we can leave here just `edit? || show_contributions?` to simplify readability.
Because show_contributions already has `show_contributions_to_everyone?` check
https://github.com/CoMakery/comakery-server/blob/8ce949517dc86aff27528bb25252a3666d1f5479/app/policies/project_policy.rb#L46-L48

Riddlerrr (commented):
I tried to review it but seems like I don't have enough context for the task.
So I'm not sure how correct it is.
Let's wait for @Alexey1100 review then

aquabu (approved):

PR #2149, opened 2021-06-01
Update/coverage score

olehyev138:
improve Total coverage score


app/jobs/ore_id_wallet_opt_in_tx_create_job.rb
```diff
@@ -25,10 +25,7 @@ def reschedule(wallet_provision)
   end
 
   def wait_to_perform(wallet_provision)
-    if wallet_provision.next_sync_allowed_after < Time.current
-      0
-    else
-      wallet_provision.next_sync_allowed_after - Time.current
-    end
+    wait_time = wallet_provision.next_sync_allowed_after - Time.current
```
aquabu, line 28: @olehyev138 are there tests for each of the logic branches here? If not tests should be added to get to full coverage.

olehyev138: @aquabu
it's working well like the previous tests. I only change the `if-else` to `ternary operator`

aquabu: @olehyev138 hmm, seems like it shouldn't be working because: 
* `Time.current.class` returns `ActiveSupport::TimeWithZone`
* `Time.current.positive?` raises `NoMethodError: undefined method `positive?'`
* In this case it is calling `positive` without the question mark which wouldn't even be defined on `Float` or `Integer` if the type was coerced to one of those.

This makes me wonder if there is a test because if there was one it seems like it should be failing. 

Could you either add a test or provide a link to the test. Thanks!

aquabu (changes requested):

app/jobs/ore_id_wallet_opt_in_tx_create_job.rb
olehyev138: I'll make new PR for this

aquabu (approved):

PR #2161, opened 2021-06-03
Blockchains: Introduce a human_url for every token type

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/178387477

image
image
image


zhinfuhnya (approved):

app/decorators/project_decorator.rb
```diff
@@ -139,7 +139,7 @@ def token_props
         symbol: token.symbol,
         network: token.blockchain.name,
         address: (h.middle_truncate(token.contract_address) if token.contract_address),
-        address_url: (token.blockchain.url_for_address_human(token.contract_address) if token.contract_address),
+        address_url: (token.blockchain.url_for_token_human(token.contract_address) if token.contract_address),
```
aquabu, line 142: I may be wrong about this, but it might be better to handle this as a method on a Token decorator instead of by modifying all the blockchain template definitions.

Consider that:
* For first layer tokens like Bitcoin, ETH, Algos, etc there won't be a token contract address. This will lead confusion when implementing the template. 
* The information needed to display the token address should already be available on the token model. We use it in a lot of places.

app/models/blockchain/cardano.rb
```diff
@@ -68,6 +68,12 @@ def url_for_address_human(addr)
     "https://#{explorer_human_host}/address/#{addr}"
   end
 
+  # Token address url on block explorer website
+  # @return [String] url
+  def url_for_token_human(addr)
```
aquabu, line 73: If this does remain in the token template models instread of going in a decorator it seems like this code is repeated a lot and might be better in a base class that is overridden for exceptions.

aquabu (changes requested):

app/models/blockchain/bitcoin.rb
```diff
@@ -68,6 +68,12 @@ def url_for_address_human(addr)
     "https://#{explorer_human_host}/address/#{addr}"
   end
 
+  # Token address url on block explorer website
+  # @return [String] url
+  def url_for_token_human(addr)
```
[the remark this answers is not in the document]
Riddlerrr: It is explorer site specific thing. We have own explorer site for each blockchain so I think it is correct place to store this urls.
We will not generate this URL for non-contracts tokens:
https://github.com/CoMakery/comakery-server/blob/ba6b77a65a3d9ee5e047f145576f89f1222dceea/app/decorators/token_decorator.rb#L17-L21

app/decorators/project_decorator.rb
Riddlerrr: I extracted to token decorator

app/models/blockchain/cardano.rb
Riddlerrr: As I remember we decided to add a new method to each blockchain class and template instead of adding into the main class.
@Alexey1100 Am I correct?

Alexey1100: Hmm, true.
I think these irregularities and confusion arise here because it's not a property of Blockchain, it should be defined in TokenType.

We can add something like `TokenType#url_for_self..` which:
1) in case of native coins will return just the homepage of blockchain/coin
2) in case of ERC20 it will call `Blockchain::Ethereum#url_for_address...` via `attrs[:blockchain]` with `attrs[:contract_address]` as a param
3) for Algorand Assets it will call `Blockchain::Algorand#url_for_asset...`
4) for Algorand Apps `Blockchain::Algorand#url_for_app...`

We already have all these helpers implemented and we can add the delegation to token, so it would work like this:
`token.url_for_self_human`

Alexey1100: In short:
The url depends both on Blockchain and TokenType, but Blockchain only provides the explorer url and doesn't know anything about token types or contract addresses. TokenType knows on what blockchain it's running and it's own contract address, so the helper should be there.

Riddlerrr:
I did required changes. Ready to review again.

app/models/token_type/algorand_security_token.rb
```diff
@@ -130,4 +130,16 @@ def supports_balance?
   def blockchain_balance(wallet_address)
     contract.app_balance(wallet_address)
   end
+
+  # Token address url on block explorer website or just a link to block explorer
+  # @return [String] url
+  def human_url
+    "https://#{blockchain.explorer_human_host}/application/#{contract_address}"
```
Alexey1100, line 137: Should use `blockchain.url_for_app_human`.
I think we only have `blockchain.url_for_app_api` since app support for explorer has been added just recently, so we need to add that.

app/models/token_type/asa.rb
```diff
@@ -112,4 +112,16 @@ def supports_balance?
   def blockchain_balance(wallet_address)
     contract.asset_balance(wallet_address)
   end
+
+  # Token address url on block explorer website or just a link to block explorer
+  # @return [String] url
+  def human_url
+    "https://#{blockchain.explorer_human_host}/asset/#{contract_address}"
```
Alexey1100, line 119: Should use `blockchain.url_for_asset_human`.

lib/generators/token_type/templates/token_type_spec.erb
```diff
@@ -20,4 +20,13 @@ describe TokenType::<%= class_name %> do
   specify { expect(described_class.new.supports_token_freeze?).to be_falsey }
   specify { expect(described_class.new.supports_balance?).to be_falsey }
   specify { expect(described_class.new.blockchain_balance('')).to raise_error(NotImplementedError) }
+
+  describe 'human url' do
+    let(:attrs) { { contract_address: nil, blockchain: Blockchain::Blockchain.new } }
```
Alexey1100, line 25: Looks like this spec template will be failing by default, since `Blockchain::Blockchain.new` is not defined.

Probably better to mock `Blockchain#explorer_host`

app/decorators/project_decorator.rb
```diff
@@ -138,8 +139,8 @@ def token_props
         name: token.name,
         symbol: token.symbol,
         network: token.blockchain.name,
-        address: (h.middle_truncate(token.contract_address) if token.contract_address),
-        address_url: (token.blockchain.url_for_address_human(token.contract_address) if token.contract_address),
+        address: helpers.middle_truncate(token.token_type.human_url_name),
```
Alexey1100, line 142: Probably good to add a feature spec, to keep track that the link is present and working.

aquabu (approved):

PR #2180, opened 2021-06-10
Invite admins to projects

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/177788724
https://www.pivotaltracker.com/n/projects/2167210/stories/178418716

**EXAMPLES CASES**

1. Email is invalid

Screen Shot 2021-06-15 at 1 58 22 PM

2. Send invite to unregistered account

Screen Shot 2021-06-15 at 1 59 03 PM

Screen Shot 2021-06-15 at 1 58 52 PM

Screen Shot 2021-06-15 at 1 59 16 PM

Screen Shot 2021-06-15 at 2 00 22 PM

Screen Shot 2021-06-15 at 2 01 55 PM

Screen Shot 2021-06-15 at 2 03 05 PM

3. Resend invitation

Screen Shot 2021-06-15 at 1 53 16 PM

4. Send invite to registered user

Screen Shot 2021-06-15 at 2 08 39 PM

Screen Shot 2021-06-15 at 2 09 00 PM

Screen Shot 2021-06-15 at 2 08 49 PM


db/schema.rb
```diff
@@ -10,9 +10,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_05_31_030108) do
+ActiveRecord::Schema.define(version: 2021_06_10_104821) do
 
   # These are extensions that must be enabled in order to support this database
+  enable_extension "pg_stat_statements"
```
Riddlerrr, line 16: Extra line was added

db/schema.rb
```diff
@@ -155,9 +156,9 @@
   end
 
   create_table "api_request_logs", force: :cascade do |t|
-    t.jsonb "body", null: false
     t.inet "ip", null: false
     t.string "signature", null: false
+    t.jsonb "body", null: false
```
Riddlerrr, line 161: This line was moved

app/interactors/send_invite.rb
```diff
@@ -1,21 +1,70 @@
 class SendInvite
   include Interactor
 
-  def call
-    context.fail!(errors: ['The user must have signed up to add them']) if account.blank?
+  delegate :params, :whitelabel_mission, :project, to: :context
 
-    invite = context.project.project_roles.new(account: account, role: context.params[:role])
+  # TODO: Refactor cyclomatic complexity
```
aquabu, line 6: good to refactor this bit

aquabu (approved):

app/controllers/accounts_controller.rb
```diff
@@ -60,7 +61,11 @@ def create
     if recaptcha_valid?(model: @account, action: 'registration') && ImagePixelValidator.new(@account, account_params).valid? && @account.save
       session[:account_id] = @account.id
 
-      UserMailer.with(whitelabel_mission: @whitelabel_mission).confirm_email(@account).deliver
+      if @project_invite.present?
+        @account.confirm!
```
Alexey1100, line 65: I don't think we should automatically confirm the email if there's a valid invitation token present.

Imagine a case:
1) You creating an invitation, leaving it unused.
2) Now you can create as many confirmed account as you need with any email addresses, just by applying token as a param to sign up url.

Alexey1100: Essentially on this step we need to check that provided email matches the project invite email and not allow to register otherwise.

app/models/invite.rb
```diff
@@ -0,0 +1,9 @@
+class Invite < ApplicationRecord
+  has_secure_token
+
+  belongs_to :invitable, polymorphic: true
+
+  validates :email, uniqueness: { case_sensitive: false }
```
Alexey1100, line 6: Should be `uniqueness: { scope: %i[invitable_id invitable_type] }`

Otherwise it's impossible to invite the account, when there are other invites (eg from different projects) pending to that email.

Alexey1100: And we need to think about the case when there are two invites sent to the same unregistered email.
If user accepts the first invite, the second one would be impossible to accept.

app/views/user_mailer/send_invite_to_platform.html.rb
```diff
@@ -0,0 +1,12 @@
+class Views::UserMailer::SendInviteToPlatform < Views::Base
```
Alexey1100, line 1: I don't think we want to use Fortitude anymore, better switch to ERB ( :

app/views/user_mailer/send_invite_to_project.html.rb
```diff
@@ -0,0 +1,10 @@
+class Views::UserMailer::SendInviteToProject < Views::Base
```
Alexey1100, line 1: Fortitude.

spec/controllers/projects/invites_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-RSpec.describe Projects::InvitesController, type: :controller do
+RSpec.describe Projects::InvitesController, type: :controller, skip: true do
```
Alexey1100, line 3: Hmm this shouldn't be skipped right?

db/migrate/20210610104821_create_invites.rb
```diff
@@ -0,0 +1,14 @@
+class CreateInvites < ActiveRecord::Migration[6.0]
+  def change
+    create_table :invites do |t|
+      t.string :email, null: false, index: { unique: true }
+      t.string :token, null: false, index: { unique: true }
+      t.string :role, null: false
```
Alexey1100, line 6: I think logically an invitable here should be ProjectRole rather than the Project.
This will also remove the need for `:role` column and make the Invite model fully reusable.

The only update is to make ProjectRole#account optional if there is a pending invite associated with it.

app/models/invite.rb
Alexey1100: ^^^^ that would probably be the case for Republic with separate lockup and non-lockup projects.

PR #2181, opened 2021-06-11
Awards coverage

olehyev138:
For Awards_controllers test
Removed some redundant code from tests, that probably served as workaround for some cases to make tests green.
Also replaced mom with factories. 
improved awards#send_award action


spec/controllers/awards_controller_spec.rb
```diff
@@ -217,6 +211,8 @@
       }
     end
 
+    before(:all) { Timecop.freeze(Time.zone.local(2021, 6, 3, 15)) }
```
aquabu, line 214: @olehyev138 I think this should probably be `before(:each)`

As you may know `before(:all)` can cause subtle side effect issues because database values and instance variables aren't reset between tests. If there isn't a strong reason to use `before(:all)` I think it's better to use `before(:each)` in this case.

aquabu (changes requested):

spec/controllers/awards_controller_spec.rb
olehyev138: that is the only workaround for that case, because we have to test form properties assignment - that correct data is set up for a form on frontend part. part of that data is an "updated_at" timestamp - I don't know why would they ever need that on a form, but i am only improving coverage now. in this project many entities are associated with each other, so when one entity gets updated through model callbacks, the associated record gets "updated_at" changed because of "tough: true" option of the corresponding association.

in this test we had to freeze time with Timecop, so that we could test exact values of "created_at" and "updated_at". but the problem is that Timecop freezes time only after all records are created via factories.
so i had to make Timecop freeze time BEFORE records are created.

"before(:all)" does exactly what I needed
no worries about unexpected behavior and incorrect reset of database records, as in this case "before(:all)" is isolated in particular context
so it works only in that particular scope

olehyev138: another way out is to remove timestamps from the form props assignment.
but that is not the scope of the ticket i am working on right now.
and that is basically a logic change.
it can impact the frontend part. i am not sure why they need those timestamps on a form.
they just decided to use kind of logic, that is hardly can be tested.

olehyev138: before(:each) works before the particular test, and after records are created. so it is not helpful in this case

btw, "before(:all)" is used in one more place
spec/lib/ecies/crypt_spec.rb:41

aquabu: @olehyev138 thanks. Maybe @Alexey1100 or @Riddlerrr knows about why the props assignment is this way and if it should be changed. Something seems a little odd and it's probably worth digging into a bit.

aquabu (approved):

aquabu (approved):

PR #2182, opened 2021-06-14
To process uploaded image before save

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/178418604

DONE:
- validate if it an correct jpg or png image
- validate image dimensions (max 4096x4096 pixels)
- strip EXIF data for jpg images
- Add optional resizing an image before save. Applied for an avatar upload
- removed fastimage gem dependency

TODO:
- [x] Add size validator before processing
- [x] Change user provided filename to auto-generated one


app/controllers/accounts_controller.rb
```diff
@@ -237,14 +235,11 @@ def project_decorate(project, award = nil)
 
     def account_decorate(account)
       account.as_json(only: %i[email first_name last_name nickname date_of_birth country ethereum_auth_address]).merge(
-        image_url: account_image_url(account)
+        image_url: helpers.account_image_url(account, 190)
       )
     end
 
-    def account_image_url(account)
```
Riddlerrr, line 244: we already have helper with the same name doing the same.

app/services/image_preparer.rb
```diff
@@ -0,0 +1,94 @@
+class ImagePreparer
+  MAX_WIDHT = 4096
+  MAX_HEIGHT = 4096
+
+  attr_accessor :record
+  attr_reader   :params, :valid, :options
+  alias valid? valid
```
zhinfuhnya, line 7: As for me it would be good to create a public method `valid?` with a logic of `validate_and_prepare_attachments` method instead of assigning a `@valid` variable and then call `validate_and_prepare_attachments`

app/services/image_preparer.rb
```diff
@@ -0,0 +1,94 @@
+class ImagePreparer
+  MAX_WIDHT = 4096
+  MAX_HEIGHT = 4096
+
+  attr_accessor :record
```
zhinfuhnya, line 5: Better to move to private scope

zhinfuhnya:
@Riddlerrr added a few non-critical notices, overall it looks super! 🚀

app/services/image_preparer.rb
Riddlerrr: Yeah, I agree 👍

app/services/image_preparer.rb
Riddlerrr: Moved it to `attr_reader`

Alexey1100 (approved):

aquabu (approved):

PR #2192, opened 2021-06-21
Refactoring of image processing and remove image size restrictions

Riddlerrr:
task:  https://www.pivotaltracker.com/story/show/178559683

This PR completely refactors image processing logic and removes UI restrictions for image dimensions.

`has_one_attached_and_prepare_image` introduced
It support resize changes, like this:
`has_one_attached_and_prepare_image :image, resize: '190x190!'`
Resize options: https://legacy.imagemagick.org/Usage/resize/#noaspect

After set attachment to image we check it using `ImagePreparer` and reset attachment from it.

ImagePreparer changes:
- It now process only one image
- it save image in `.error` and set `attachment` to nil in case of errors


spec/models/mission_spec.rb
```diff
@@ -1,9 +1,6 @@
 require 'rails_helper'
-require 'models/concerns/active_storage_validator_spec'
 
 describe Mission do
-  it_behaves_like 'active_storage_validator', %w[image logo whitelabel_logo whitelabel_logo_dark whitelabel_favicon]
```
Alexey1100, line 5: I think we need to implement ImagePreparer spec as shared example, and then calling it in model spec defining what fields should be processed, similar as we had before with ActiveStorageValidator.

This way we can be sure that the attachments set correctly on model level, having 100% coverage.

app/models/application_record.rb
```diff
@@ -1,3 +1,16 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
+
+  def save_if_no_errors
```
Alexey1100, line 4: Hmm what's the purpose of this method?

`#save` and `#update` will return false if there are any errors.

app/models/concerns/prepare_image.rb
```diff
@@ -0,0 +1,34 @@
+module PrepareImage
+  extend ActiveSupport::Concern
+
+  class_methods do
+    # rubocop:todo Naming/PredicateName
+    def has_one_attached_and_prepare_image(name, **options)
+      has_one_attached :"#{name}"
+
+      generated_association_methods.class_eval <<-CODE, __FILE__, __LINE__ + 1
+        def #{name}=(attachable)
```
Alexey1100, line 10: Is this part covered with specs?
I think a shared example would be helpful.

Alexey1100 (changes requested):

app/models/concerns/prepare_image.rb
aquabu: +1

app/models/application_record.rb
Riddlerrr: It's a hack.
We can add an error during image processing but if we call regular `#save` it will clear all errors and do validation from scratch. But we want to show that image error instead.

app/models/project.rb
```diff
@@ -3,17 +3,14 @@
...
-  # attachment :square_image, type: :image
-  # attachment :panoramic_image, type: :image
-  has_one_attached :image
-  has_one_attached :square_image
-  has_one_attached :panoramic_image
-  has_one_attached :transfers_csv
+  has_one_attached_and_prepare_image :image
+  has_one_attached_and_prepare_image :square_image
+  has_one_attached_and_prepare_image :panoramic_image
+  has_one_attached_and_prepare_image :transfers_csv
```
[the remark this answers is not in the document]
Riddlerrr: mindfulness 👍

app/models/application_record.rb
Riddlerrr: I thought about it more and came up with better approach. Now it close to perfect =)

I save an error in `uploaded_image_errors` variable and set it in `before_validation` callback, so we can use usual `save` and `update` methods

Alexey1100 (approved):

aquabu (approved):

PR #2197, opened 2021-06-26
Coverage controllers projects

olehyev138:
Improve projects controllers coverage score to 100%

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -63,7 +63,7 @@ class ProjectsController < ApplicationController
 
     authorize @project
 
-    if @project.save
+    if ImagePreparer.new(@project, project_params).valid? && @project.save
       set_generic_props
       camelize_props
       render json: { id: @project.id, props: @props }, status: :ok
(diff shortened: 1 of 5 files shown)


app/controllers/projects_controller.rb
```diff
@@ -309,10 +311,6 @@ def set_award
       @can_award = awardable_types_result.can_award
     end
 
-    def project_detail_path
```
aquabu, line 312: @olehyev138 just checking why this is deleted? Is it no longer used?

app/controllers/projects_controller.rb
```diff
@@ -349,25 +347,12 @@ def project_props(project)
...
     def token_props(token)
-      if token.present?
-        token.as_json(only: %i[name symbol _token_type]).merge(
-          image_url: GetImageVariantPath.call(attachment: token.logo_image, resize_to_fill: [25, 18]).path
-        )
-      end
-    end
+      return if token.nil?
 
-    def mission_props(mission)
```
aquabu, line 359: Is this also no longer used?

aquabu (changes requested):

app/controllers/projects_controller.rb
olehyev138: this is not used too

olehyev138: @aquabu they are not used and so I removed theme instead add tests.

aquabu (approved):

PR #2199, opened 2021-06-28
Allow an observer to export transfers 

Alexey1100:
Resolves:
https://www.pivotaltracker.com/story/show/178677002

### DIFF
--- a/spec/policies/project_policy_spec.rb
+++ b/spec/policies/project_policy_spec.rb
@@ -158,26 +158,17 @@ describe ProjectPolicy do
     end
   end
 
-  describe '#export_transfers?' do
-    subject { described_class.new(user, my_public_project).export_transfers? }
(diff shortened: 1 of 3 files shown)


spec/policies/project_policy_spec.rb
```diff
@@ -158,6 +158,28 @@
...
 
+  describe '#export_transfers?' do
+    subject { described_class.new(user, my_public_project).export_transfers? }
+
+    context 'without user' do
+      let(:user) { nil }
+
+      it { is_expected.to be_falsey }
+    end
+
```
aquabu, line 169: @Alexey1100 I think this needs a test that any other roles like `member` are falsey

aquabu (changes requested):

zhinfuhnya (approved):

aquabu (approved):

PR #2202, opened 2021-06-29
models/project coverage score improvment

olehyev138:
app/models/project.rb

Coverage score improvement


Alexey1100:
Looks like resolved by #2245

PR #2204, opened 2021-06-30
Creating new user or editing permission of the existing user breaks the Accounts table design

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/178614552

- Added three dots menu display regardless of the user's role
- Added lazy loading for settings
- Added blocking menu settings (if the user has no permissions)
- Removed `app/javascript/controllers/accept_turbo_stream_controller.js `


Riddlerrr (approved):
👍

app/controllers/projects/accounts/settings_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Projects::Accounts::SettingsController < ApplicationController
+  skip_after_action :verify_authorized, :verify_policy_scoped
+
+  def show
+    @project_role = ProjectRole.find_by(project_id: params[:project_id], account_id: params[:account_id])
```
Alexey1100, line 5:
```suggestion
    @project_role = ProjectRole.find_by!(project_id: params[:project_id], account_id: params[:account_id])
```

Otherwise it may raise 500 on next line.

app/controllers/projects/accounts/settings_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Projects::Accounts::SettingsController < ApplicationController
+  skip_after_action :verify_authorized, :verify_policy_scoped
+
+  def show
```
Alexey1100, line 4: We need to add policy here, otherwise any logged in user can see an account role when following the link, even when not having access to the project.

```ruby
authorize @project, :accounts?
```

Alexey1100 (changes requested):

app/controllers/projects/accounts/settings_controller.rb
Alexey1100: Ahh okay, we are not displaying role name in the template, only id.
Nevertheless worth adding the policy check.

Alexey1100 (changes requested):
Still will raise 500 in template if project role is not found :)

Alexey1100 (approved):

PR #2205, opened 2021-06-30
Fix specs for acceptance branch

zhinfuhnya:
Fixes specs for acceptance branch.

### DIFF
--- a/spec/controllers/projects_controller_spec.rb
+++ b/spec/controllers/projects_controller_spec.rb
@@ -729,26 +729,6 @@ describe ProjectsController do
           end
         end
       end
-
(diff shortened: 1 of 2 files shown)


spec/controllers/projects_controller_spec.rb
```diff
@@ -729,26 +729,6 @@
           end
         end
       end
-
-      context 'when with invitation' do
```
Riddlerrr, line 733: This was moved to Inviration controller in https://github.com/CoMakery/comakery-server/pull/2195
So no need anymore

aquabu (approved):

PR #2209, opened 2021-07-01
Increase invite token length

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/178703558


app/models/invite.rb
```diff
@@ -1,4 +1,6 @@
 class Invite < ApplicationRecord
+  TOKEN_LENGTH = 72
+
   has_secure_token
```
Alexey1100, line 4: I think it should be:

```ruby
has_secure_token :token, length: TOKEN_LENGTH
```

app/models/invite.rb
```diff
@@ -14,6 +16,10 @@ class Invite < ApplicationRecord
 
   scope :pending, -> { where(accepted: false) }
 
+  def self.generate_unique_secure_token
```
Alexey1100, line 19: Looks like we don't need this method.

spec/models/invite_spec.rb
```diff
@@ -45,7 +45,7 @@
     end
   end
 
-  describe '.pending' do
+  describe '#pending' do
```
Alexey1100, line 48: Pending is a class method :)

Alexey1100 (changes requested):

app/models/invite.rb
zhinfuhnya: according to the documentation it should work but in practice I get an error:
`ArgumentError (wrong number of arguments (given 2, expected 0..1))`

app/models/invite.rb
zhinfuhnya: So, it's why I use this trick

app/models/invite.rb
Alexey1100: Ahh okay, looks like it was added in rails 6.1.
https://github.com/rails/rails/commit/51443e22f3cba76907204cc528ebab43224adfd9

We should probably upgrade since it was released more than a half of year ago, but yeah it's out of scope for this PR.

app/models/invite.rb
Alexey1100: Okay, maybe better to update the spec to verify that the `invite#token` is actually populated on creation with the right value, and add a comment that it's an override of default method, because it's a bit unclear now.

Alexey1100 (approved):

PR #2210, opened 2021-07-01
Check confidentiality for whitelabel

sofiyareverse:
https://www.pivotaltracker.com/n/projects/2167210/stories/178387431

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -136,7 +136,6 @@ class ProjectsController < ApplicationController
       @q = policy_scope(@project_scope).ransack(params[:q])
       @q.sorts = 'project_roles_count DESC' if @q.sorts.empty?
 
-      # check if confidentiality
       @projects_all = @q.result.with_all_attached_images.includes(:token, :mission, :project_admins, account: [image_attachment: :blob])
       @projects = @projects_all.page(@page).per(9)
 
@@ -356,7 +355,7 @@ class ProjectsController < ApplicationController
     end
 
     def redirect_for_whitelabel
-      if @whitelabel_mission && !@project.require_confidentiality?
+      if @whitelabel_mission
         if policy(@project).show_contributions?
           redirect_to project_dashboard_transfers_path(@project)
         else
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -136,6 +136,7 @@ class ProjectsController < ApplicationController
       @q = policy_scope(@project_scope).ransack(params[:q])
       @q.sorts = 'project_roles_count DESC' if @q.sorts.empty?
 
+      # check if confidentiality
       @projects_all = @q.result.with_all_attached_images.includes(:token, :mission, :project_admins, account: [image_attachment: :blob])
       @projects = @projects_all.page(@page).per(9)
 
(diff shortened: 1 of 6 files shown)


app/controllers/projects_controller.rb
```diff
@@ -355,7 +356,7 @@ def token_props(token)
     end
 
     def redirect_for_whitelabel
-      if @whitelabel_mission
+      if @whitelabel_mission && !@project.require_confidentiality?
```
zhinfuhnya, line 359: this condition already applies to `policy(@project).show_contributions?`
so, it's unnecessary double check

look into `app/policies/project_policy.rb`

method `show_contributions?` call `show_contributions_to_everyone?` which check:
`project.public? && !project.require_confidentiality?`

app/models/project.rb
```diff
@@ -80,6 +80,7 @@ class Project < ApplicationRecord
   scope :unarchived, -> { where.not visibility: 4 }
   scope :publics, -> { where 'projects.visibility in(1)' }
   scope :with_all_attached_images, -> { with_attached_image.with_attached_square_image.with_attached_panoramic_image }
+  scope :confidential, -> { where(require_confidentiality: true) }
```
zhinfuhnya, line 83: where does this scope used?

sofiyareverse: it is a draft. I will continue with it today

app/policies/project_policy.rb
```diff
@@ -18,7 +18,7 @@ def resolve
       if @account
         account.accessable_projects(scope)
       else
-        scope.publics
+        scope.publics.non_confidential
```
Alexey1100, line 21: Hmm, we should include confidential projects if they are public.
Confidentiality flag just hides transfers/accounts pages, but project overview and tasks/roadmap are still visible.

I guess the issue is that these visible pages are disabled for whitelabel projects, so enabling confidentiality on whitelabel is pretty much equivalent of making the project non-public.

So, maybe we should apply the scope above only for the whitelabel requests.

sofiyareverse:
According to our last conversation I checked the question about request as param in policy scope and have found the next answer: in application_controller method "set_project_scope" we return whitelabels for current domain, so to my mind we shouldn't check request again in the projects_controller method set_project

spec/policies/project_policy_spec.rb
```diff
@@ -58,8 +58,28 @@
...
+        let!(:my_whitelabel_confidential_project) { create(:project, title: 'whitelabel confidential', account: project_account, visibility: 'public_listed', require_confidentiality: true, mission: create(:mission, whitelabel: true)) }
+
+        it 'returns all public non-confidential projects for whitelabels if account is nil' do
+          whitelabel_projects = Project.where(whitelabel: true)
+          projects = ProjectPolicy::Scope.new(nil, whitelabel_projects).resolve
+
+          expect(projects.map(&:title).sort).to eq([my_whitelabel_non_confidential_project].map(&:title).sort)
+        end
+
+        it 'returns all public confidential and non-confidential projects for whitelabels if account is nil' do
```
Alexey1100, line 76:
```suggestion
        it 'returns all public confidential and non-confidential projects for whitelabels if account has access' do
```

Alexey1100 (approved):

PR #2212, opened 2021-07-02
Grant project member role to transfer recipient

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/178716195


spec/policies/project_policy_spec.rb
```diff
@@ -163,7 +163,6 @@
 
     it { is_expected.to permit(project_account, my_public_project) }
     it { is_expected.to permit(project_admin, my_public_project) }
-    it { is_expected.to permit(project_contributor, my_public_project) }
```
Alexey1100, line 166: Let's move it to `.not_to permit` section.

zhinfuhnya:
We have a very strange type comparison, for example we have an expected object with nested `instances_of(Type)` and this throws an error when compared to the real object. To summarize, type checking is possible only in `expect` scope.

Here are examples of the fallen spec:

image

aquabu (approved):

PR #2216, opened 2021-07-02
Disable Rspec retries in CI

Alexey1100:
Resolves:
https://www.pivotaltracker.com/story/show/178677287

### DIFF
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -137,7 +137,7 @@ jobs:
       run: env bundle exec rails db:migrate
 
     - name: Run tests
-      run: env bundle exec rake knapsack_pro:queue:rspec || env bundle exec rspec --only-failures --format documentation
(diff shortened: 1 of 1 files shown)


Alexey1100:
I've disabled test retry, trying to avoid simplecov issue, since the retry run probably overwrites the simplecov report for the full run, making it incomplete.

But looks like we have lots of flaky tests, related to API ids stubbing.

aquabu:
good catch @Alexey1100

Riddlerrr:
@Alexey1100 
As a workaround, what if we force disable `SIMPLECOV_ENABLED` for retry failures:
```
SIMPLECOV_ENABLED=false env bundle exec rspec --only-failures --format documentation
```

This way simplecov will not overwrite results on second rspec run.

Alexey1100:
Resolved in #2233

PR #2218, opened 2021-07-05
HW: Show heroku deploy buttons for algorand tokens

Riddlerrr:
tasks: 
- https://www.pivotaltracker.com/story/show/178366933
- https://www.pivotaltracker.com/story/show/178588511

To auto-fill `PURESTAKE_API` it must be present in server ENV vars
Deploy button enabled for algorand security token projects only

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -75,8 +75,7 @@ module ApplicationHelper
     }
 
     if token._token_type_on_ethereum?
-      token_type = token._token_type
-      token_type = 'erc20_batch' if token_type == 'erc20' && token.batch_contract_address.present?
+      token_type = ethereum_token_type(token)
       approval_address =
         case token_type
         when 'erc20_batch' then token.batch_contract_address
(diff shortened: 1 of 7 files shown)


zhinfuhnya (approved):

app/helpers/application_helper.rb
```diff
@@ -88,8 +87,17 @@ def deploy_to_heroku_url(project)
       params[:env][:ETHEREUM_TOKEN_SYMBOL] = token.symbol
       params[:env][:ETHEREUM_CONTRACT_ADDRESS] = token.contract_address
       params[:env][:ETHEREUM_APPROVAL_CONTRACT_ADDRESS] = approval_address if token_type.in?(%w[erc20_batch token_release_schedule])
+    elsif token._token_type_on_algorand?
+      params[:env][:PURESTAKE_API] = ENV['PURESTAKE_API']
+      params[:env][:OPT_IN_APP] = token.contract_address
```
Alexey1100, line 92: Seems like lots of conditions here are based on token type.
So, it would be logically to move the hw deploy url into token type subclass.

Alexey1100 (approved):

app/helpers/application_helper.rb
Riddlerrr: But it still require to fill vars from `request` and from `project`, so not sure it will be easier

Alexey1100: Yep, we can pass project id and comakery base url as parameters to the method.

PR #2226, opened 2021-07-06
Wallet changes for nil blockchain

sofiyareverse:
https://www.pivotaltracker.com/n/projects/2167210/stories/177956343

You can be with "blockchain: nil" all this steps. If you are not ore_id_account you will have this error: 
"name": [
                "can't be blank"
            ]
For ore_id_account you will have this error: 
"blockchain": [
                "unknown blockchain value",
                "is not supported with ore_id source"
            ],
            "name": [
                "can't be blank"
            ]
Steps that can be made before getting errors stack:
wallets_controller.rb -> action create -> wallets_params -> wallet_creator.rb -> wallet_creator/provision.rb -> wallet.rb (validations)


Riddlerrr:
@sofiyareverse  Blockchain is mandatory field for an Wallet and there are no Wallets without blockchain filled on demo:
```
> Wallet.where(_blockchain: nil).count
=> 0
```

So I think the issue should be somewhere else

Alexey1100:
Yep, the error is raised when we trying to create a wallet.
I remember that at some point we are converting user provided param `blockchain` into internal enum `_blockchain`, so probably in some cases validation is triggered before we perform the conversion.

spec/acceptance/api/v1/19_wallets_spec.rb
```diff
@@ -131,6 +131,28 @@
...
+      let(:ast_token) { create(:algo_sec_token, id: 55) }
+      let(:reg_group) { create(:reg_group, id: 45, token: ast_token) }
+      let(:tokens_to_provision) do
+        [
+          { token_id: asa_token.id.to_s },
+          { token_id: ast_token.id.to_s, max_balance: '100', lockup_until: '1', reg_group_id: reg_group.id.to_s, account_frozen: 'false' }
+        ]
+      end
+
+      example 'CREATE WALLET – ORE_ID WITH PROVISIONING – ERROR' do
```
Riddlerrr, line 147: These acceptance tests need for docs generation, so if you don't want to show something in docs it should be placed in usual controller tests.
I think we already have example with an error, so let's move into controller tests.

Riddlerrr (changes requested):

spec/controllers/api/v1/wallets_controller_spec.rb
```diff
@@ -80,6 +80,17 @@
         expect(response).to have_http_status(400)
         expect(assigns[:errors][0][:_blockchain]).to eq ['unknown blockchain value']
       end
+
+      it 'creates a provisioned wallet' do
+        token = create(:asa_token)
```
zhinfuhnya, line 85: It's a bad practice to create examples inside `it` scope.

Riddlerrr (approved):

PR #2227, opened 2021-07-06
Project tasks visibility

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/178336397


app/controllers/award_types_controller.rb
```diff
@@ -1,16 +1,19 @@
 class AwardTypesController < ApplicationController
+  rescue_from Pundit::NotAuthorizedError, with: :not_authorized
```
Alexey1100, line 2: We have the same code in ApplicationController, which should be inherited here.
Is it not working?

app/javascript/components/layouts/Header.jsx
```diff
@@ -80,7 +80,7 @@ const navContent = (isLoggedIn, isAdmin, isWhitelabel, currentPath) => {
           My Projects
         </a>
 
-        { !isWhitelabel &&
+        { projectAwardsVisible &&
```
Alexey1100, line 83: Should always display on non-whitelabel.

app/policies/award_policy.rb
```diff
@@ -18,10 +24,8 @@ def resolve
     end
   end
 
-  def initialize(account, award)
-    @account = account
-    @award = award
-    @project = @award.project
+  def index?
+    @project.mission&.project_awards_visible?
```
Alexey1100, line 28: Will not work properly in non-whitelabel env.

app/policies/award_type_policy.rb
```diff
@@ -18,4 +24,8 @@ def resolve
       end
     end
   end
+
+  def index?
+    @project.mission&.project_awards_visible?
```
Alexey1100, line 29: Will not work properly in non-whitelabel env.

app/policies/award_policy.rb
Alexey1100: We should always display the page there, regardless mission.

app/policies/project_policy.rb
```diff
@@ -52,7 +52,7 @@ def show_contributions?
   def show_award_types?
     return false if ENV['WHITELABEL'] == true
 
-    (show? || unlisted?)
+    (show? || unlisted?) && project.mission&.project_awards_visible?
```
Alexey1100, line 55: Will not work properly in non-whitelabel env.

db/migrate/20210706104324_add_project_awards_visible_to_missions.rb
```diff
@@ -0,0 +1,5 @@
+class AddProjectAwardsVisibleToMissions < ActiveRecord::Migration[6.0]
+  def change
+    add_column :missions, :project_awards_visible, :boolean, default: false
```
Alexey1100, line 3: Needs `null: false`

app/controllers/award_types_controller.rb
```diff
@@ -2,15 +2,18 @@ class AwardTypesController < ApplicationController
...
   before_action :set_award_types, only: %i[index]
   before_action :set_form_props, only: %i[new edit]
   before_action :set_index_props, only: %i[index]
   skip_after_action :verify_policy_scoped, only: %i[index]
   skip_before_action :require_login, only: %i[index]
 
-  def index; end
+  def index
+    authorize @project, :show_award_types?
+  rescue Pundit::NotAuthorizedError
```
Alexey1100, line 14: Still not sure why we are adding it here, instead of relying on default auth rescue in ApplicationController : )

spec/features/award_types/index_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    context 'with visible awards' do
+      let(:mission) { create :mission, project_awards_visible: true }
+
+      it 'shows award types page' do
+        expect(page).to have_current_path(project_award_types_path(project))
+      end
+    end
+
+    context 'with hidden awards' do
+      let(:mission) { create :mission, project_awards_visible: false }
```
Alexey1100, line 20: This logic should only be triggered with a whitelabel mission.

Alexey1100 (changes requested):
According to ticket the logic should only apply to whitelabel env.

In the non-whitelabel env the page should be accessible regardless the project mission settings, also given that the mission there is optional.

app/javascript/components/layouts/Header.jsx
zhinfuhnya: So, if the mission isn't a `whitelabel` we shouldn't display `My Tasks`?

app/policies/award_policy.rb
zhinfuhnya: Why it shouldn't work in the case of a non-whitelabel mission?

app/policies/award_type_policy.rb
zhinfuhnya: https://github.com/CoMakery/comakery-server/pull/2227#discussion_r664901098

app/policies/project_policy.rb
zhinfuhnya: https://github.com/CoMakery/comakery-server/pull/2227#discussion_r664901098

app/controllers/award_types_controller.rb
zhinfuhnya: Regarding to requirements:

`Then I should not see "My Tasks" in the site navigation
And when I visit /tasks then I should be redirected to /projects
When I visit a project at /projects/:id
And I should not see the project "tasks" menu
And when I visit /projects/:id/batches then I should be redirected to the 404 error page
`

We must have unique fallback paths for each case. We can't rely on default rescue path (`root_path`).

spec/features/award_types/index_spec.rb
zhinfuhnya: Thus we consider that the changes will only be relative to whitelabel mission (in the case of a non-whitelabel we always display tasks)?

app/policies/award_policy.rb
Alexey1100: Mission is not required for non-whitelabel projects, so in this case it will return false and page would not be displayed.

app/controllers/award_types_controller.rb
Alexey1100: I think it's typo :)
We have default redirect to root when not-authorised

cc @aquabu

Alexey1100: Considering it's a fail-state and user shouldn't be ever following this url, I think it's better to just rely on default behaviour across the website.

app/javascript/components/layouts/ProjectSetupHeader.jsx
```diff
@@ -177,7 +177,8 @@ class ProjectSetupHeader extends React.Component {
                 GitHub
               </NavLink>
               }
-              {(owner || observer || project.showBatches) && !project.whitelabel && !project.tokenLockup &&
+              {((owner || observer || project.showBatches) && !project.whitelabel && !project.tokenLockup ||
+                (project.whitelabel && project.projectAwardsVisible)) &&
```
Alexey1100, line 181: Looks like something is wrong here.

We are checking that project **is not whitelabel** AND (token is not lockup OR project **is whitelabel** ...)

config/database.yml
```diff
@@ -35,9 +35,9 @@ default: &default
 
 development:
   <<: *default
-  database: <%= ENV["DATABASE"] || Rails.application.config.project_slug + '_development_' + branch %>
```
Alexey1100, line 38: This shouldn't be checked in :)

app/controllers/award_types_controller.rb
```diff
@@ -2,15 +2,17 @@ class AwardTypesController < ApplicationController
...
   before_action :set_award_type, only: %i[edit update destroy]
-  before_action :set_award_types, only: %i[index]
   before_action :set_form_props, only: %i[new edit]
   before_action :set_index_props, only: %i[index]
   skip_after_action :verify_policy_scoped, only: %i[index]
   skip_before_action :require_login, only: %i[index]
 
-  def index; end
+  def index
+    raise Pundit::NotAuthorizedError unless policy_scope.index?
```
Alexey1100, line 14: I think this should be:

```ruby
authorize @project, :show_award_types?
```

While project policy should be:

```ruby
def show_award_types?
  (show? || unlisted?) && (!whitelabel || mission.project_awards_visible?)
end
```

app/policies/project_policy.rb
```diff
@@ -50,9 +50,7 @@ def show_contributions?
   end
 
   def show_award_types?
-    return false if ENV['WHITELABEL'] == true
-
-    (show? || unlisted?)
+    show? || unlisted?
```
Alexey1100, line 53:
```ruby
def show_award_types?
  (show? || unlisted?) && (!whitelabel || mission.project_awards_visible?)
end
```

(See comment above)

app/controllers/awards_controller.rb
```diff
@@ -28,11 +28,19 @@ class AwardsController < ApplicationController
...
   skip_before_action :require_login, only: %i[confirm show]
   skip_after_action :verify_authorized, only: %i[confirm]
-  skip_after_action :verify_policy_scoped, only: %(index)
+  skip_after_action :verify_policy_scoped, only: [:index]
   before_action :redirect_back_to_session, only: %i[index]
   before_action :create_project_role_from_session, only: %i[index]
 
-  def index; end
+  def index
+    policy_scope = AwardPolicy::Scope.new(current_account, Award, @whitelabel_mission)
```
Alexey1100, line 36: We shouldn't use scope to authorise an action, but to define a collection based on account.

Since we don't have any particular instance here to initialise the policy, I think the solution would be something like:

```ruby
raise Pundit::NotAuthorizedError if (whitelabel_mission && !@whitelabel_mission.project_awards_visible?)
```

app/policies/award_policy.rb
```diff
@@ -16,12 +23,12 @@ def resolve
...
 
-  def initialize(account, award)
-    @account = account
-    @award = award
-    @project = @award.project
+    def index?
+      return true if whitelabel_mission.blank?
+
+      whitelabel_mission.project_awards_visible?
+    end
```
Alexey1100, line 31: Scope shouldn't be used for authorisation.
See comment above.

app/policies/award_policy.rb
```diff
@@ -1,12 +1,19 @@
...
+    @award = award
+    @project = @award.project
+  end
 
   class Scope < Scope
-    attr_reader :account, :scope
+    attr_reader :account, :scope, :whitelabel_mission
 
-    def initialize(account, scope)
+    def initialize(account, scope, whitelabel_mission = nil)
```
Alexey1100, line 13: whitelabel_mission is not needed here

app/policies/award_type_policy.rb
```diff
@@ -1,21 +1,41 @@
...
     def resolve
-      if @project_editable
+      if project_policy.edit?
         scope.all
       else
         scope.where.not(state: :draft)
       end
     end
+
+    def index?
```
Alexey1100, line 27: Same here, scope should not define auth logic.

app/policies/award_type_policy.rb
```diff
@@ -1,21 +1,41 @@
...
+    @account = account
+    @award_type = award_type
+  end
+
   class Scope < Scope
-    attr_reader :account, :scope
+    attr_reader :account, :project, :whitelabel_mission, :scope
 
-    def initialize(account, project, scope)
+    def initialize(account, project, whitelabel_mission, scope)
```
Alexey1100, line 12: It makes whitelabel_mission not needed here.

Alexey1100 (changes requested):
I think we can make it much simpler :)

app/controllers/awards_controller.rb
zhinfuhnya: Heh, regarding to previous conversation we decided auth action with scope and move `whitelabel` inside policy 🤪

https://comakery-inc.slack.com/archives/C01HNM6JYGM/p1625855765158200

app/javascript/components/layouts/ProjectSetupHeader.jsx
zhinfuhnya: The mission has a `project_awards_visible` switch, which means that if it is a whitelabel and the awards are visible, we display the link.

config/database.yml
zhinfuhnya: 🙃

app/controllers/awards_controller.rb
```diff
@@ -28,11 +28,13 @@ class AwardsController < ApplicationController
...
   skip_before_action :require_login, only: %i[confirm show]
   skip_after_action :verify_authorized, only: %i[confirm]
-  skip_after_action :verify_policy_scoped, only: %(index)
+  skip_after_action :verify_policy_scoped, only: [:index]
   before_action :redirect_back_to_session, only: %i[index]
   before_action :create_project_role_from_session, only: %i[index]
 
-  def index; end
+  def index
+    authorize(@project, :show_whitelabel_awards?) if @whitelabel_mission.present?
```
Alexey1100, line 36: Hmm, we don't have @project here.

```ruby
before_action :set_project, except: %i[index confirm]
```

It should just check whether @whitelabel_mission is present and allows the awards.

Alexey1100: Specs are passing, so looks like this part is not covered.

Alexey1100 (changes requested):

app/controllers/awards_controller.rb
zhinfuhnya: https://github.com/CoMakery/comakery-server/pull/2227/commits/0c0420b510f063e73650204d8528ed73eb8de27a

app/controllers/awards_controller.rb
```diff
@@ -1,7 +1,7 @@
 class AwardsController < ApplicationController
   before_action :set_project, except: %i[index confirm]
   before_action :unavailable_for_lockup_token, except: %i[index confirm]
-  before_action :unavailable_for_whitelabel
+  before_action :unavailable_for_whitelabel, except: [:index]
```
Alexey1100, line 4: Ohh right, looks like we need to remove this callback completely from here.
Otherwise user would not be able to start/submit a task.

Alexey1100 (changes requested):

Alexey1100 (approved):

PR #2228, opened 2021-07-06
Fix display burn transfers

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/178776784


app/controllers/dashboard/transfers_controller.rb
```diff
@@ -102,7 +102,6 @@ def query
         .completed_or_cancelled
 
       @transfers_unfiltered = @transfers_unfiltered.not_cancelled unless params.fetch(:q, {}).fetch(:filter, nil) == 'cancelled'
-      @transfers_unfiltered = @transfers_unfiltered.not_burned unless transfer_type_name == 'burn'
```
Alexey1100, line 105: Does the chart works properly after the update?
And I think we need a spec to make sure burn transfers are included in the table.

Riddlerrr:
Not related to this task, just thoughts about optimisations of `fetch_chart_data`.

If we already lazy load it, we can just send a list of transfers ids we want to show in the chart. We already know it because we did it for index page.
This way we don't have to do all these duplicated things that we do for index:
https://github.com/CoMakery/comakery-server/blob/2b694d5459ba29b2e50a10866329e6a056b50e3e/app/controllers/dashboard/transfers_controller.rb#L76-L84

@Alexey1100 @zhinfuhnya What do you guys think?

zhinfuhnya:
@Riddlerrr I totally agree that we need to rewrite this in a good way : )

I think we need create a ticket for it and deal with it after the current priority scope.

aquabu (approved):

PR #2230, opened 2021-07-07
Optimize PopulateRecipientWalletForAwards data migration

Riddlerrr:
Fixed stuck data migration on demo


Riddlerrr:
@Alexey1100 Even after the data migration we still have 12k transfers without `recipient_wallet` on demo:
```
> Award.paid.where(recipient_wallet: nil).where.not(account_id: nil).count
   (19.1ms)  SELECT COUNT(*) FROM "awards" WHERE "awards"."status" = $1 AND "awards"."recipient_wallet_id" IS NULL AND "awards"."account_id" IS NOT NULL  [["status", 5]]
=> 11989
```

zhinfuhnya (approved):

PR #2231, opened 2021-07-07
Setup HTML proofer

o-bahriy:
https://www.pivotaltracker.com/story/show/178795249


config/application.rb
```diff
@@ -60,6 +60,8 @@ def self.camelize(string)
     config.react.camelize_props = true
     config.middleware.use OliveBranch::Middleware, inflection: 'camel', camelize: CustomCamelize.method(:camelize)
 
+    config.middleware.use HTMLProofer::Middleware if Rails.env.test?
+    config.middleware.use HTMLProofer::Middleware if Rails.env.development?
```
zhinfuhnya, line 64: Better to reduce to one line:
config.middleware.use HTMLProofer::Middleware if Rails.env.development? || Rails.env.test?

: )

app/views/accounts/index.html.erb
```diff
@@ -7,27 +7,25 @@
...
+                  <th class="h10 n8-color">Address</th>
+                  <th class="h10 n8-color">Primary</th>
+                  <th class="h10 n8-color">Vetting Status</th>
+                  <th class="h10 n8-color">Vetting Date</th>
+                  <th class="fixed-side"></th>
+                </tr>
               </thead>
-              <body>
-              <%= turbo_stream_from "mission_#{@whitelabel_mission&.id}_account_wallets" %>
-              <%= turbo_frame_tag "mission_#{@whitelabel_mission&.id}_account_wallets" do %>
```
Alexey1100, line 25: Apparently, there was nothing wrong with html-proofer setup.
Html spec requires `table` to have all child elements to be inside `tr`, so I've moved turbo stream tag out of the table.

Turbo frame tag is not needed here as well, as it's only required when HTTP response triggered from the page includes turbo frames and we want to modify content wrapped in the tag. We are using turbo streams here, so no turbo frames needed.

zhinfuhnya: it is necessary to revert the turbo frame.

zhinfuhnya (changes requested):

zhinfuhnya (approved):

aquabu (approved):

PR #2247, opened 2021-07-13
Introduce NextBlockchainTransactables class and remove ready_for_blockchain_transaction

Riddlerrr:
Refactoring and implementing of https://www.pivotaltracker.com/story/show/177083158


spec/controllers/api/v1/blockchain_transactions_controller_spec.rb
```diff
@@ -155,7 +155,7 @@
     end
 
     context 'with award batch available for transaction' do
-      let!(:project) { create(:project, token: create(:lockup_token), transfer_batch_size: transfer_batch_size) }
+      let!(:project) { create(:project, token: create(:lockup_token, batch_contract_address: build(:ethereum_address_1)), transfer_batch_size: transfer_batch_size) }
```
Alexey1100, line 158: We don't need `batch_contract_address` for lockup token.
Is it not working without it?

🤔

app/queries/next_blockchain_transactables.rb
```diff
@@ -0,0 +1,143 @@
...
+      q = scopes_for_awards(q)
+    elsif transactable_class == AccountTokenRecord
+      q = scopes_for_account_token_records(q)
+    elsif transactable_class == TransferRule
+      q = scopes_for_transfer_rules(q)
+    end
+
+    q = filter_unverified_accounts(q, transactable_class) if verified_accounts_only
+    q = filter_for_batch_tx(q) if batch_size > 1
+    q.limit(batch_size)
```
Alexey1100, line 68: Looks like these lines (66-68) should be only applicable for awards.

app/queries/next_blockchain_transactables.rb
```diff
@@ -0,0 +1,143 @@
...
+      SQL
+
+      scope.joins(verified_joins_sql).where('account_verified.passed = true')
+    end
+
+    def filter_for_batch_tx(scope)
+      scope
+        .joins(:transfer_type).where.not('transfer_types.name': %w[mint burn])
+        .joins(:token).where(<<~SQL, token_release_schedule: Token._token_types[:token_release_schedule])
+          (tokens.batch_contract_address IS NOT NULL) OR
```
Alexey1100, line 113: Should verify here that token type is erc20.

app/queries/next_blockchain_transactables.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    def hot_wallet_auto?
+      target[:for] == :hot_wallet && project.hot_wallet_auto_sending?
+    end
+
+    def hot_wallet_support?(transactable_class)
+      return true if target[:for] != :hot_wallet
+
+      !project.hot_wallet_disabled? && transactable_class != TransferRule
```
Alexey1100, line 133: I think better to make it more readable, just leaving:

```ruby
def hot_wallet_support?(transactable_class)
  transactable_class != TransferRule
end
```

And inside `call` we can use:
```ruby
next if target_hot_wallet? && hot_wallet_disabled?
next if target_hot_wallet? && !hot_wallet_support?(transactable_class)
```

app/queries/next_blockchain_transactables.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    def batch_support?(transactable_class)
+      transactable_class == Award && project.transfer_batch_size > 1 && project.token.batch_contract_address.present?
```
Alexey1100, line 137: Or token type is `:token_release_schedule`

spec/queries/next_blockchain_transactables_spec.rb
```diff
@@ -0,0 +1,92 @@
+require 'rails_helper'
+
+describe NextBlockchainTransactables do
+  describe '#call' do
+    let(:call) { described_class.new(project: project, target: { for: :hot_wallet }, verified_accounts_only: false).call }
```
Alexey1100, line 5: Looks like this line is not used

spec/queries/next_blockchain_transactables_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+        it 'returns the first transfer' do
+          is_expected.to eq [award1]
+        end
+
+        context 'with batch enabled' do
+          let(:batch_size) { 100 }
+
+          it 'returns all exist transfers' do
+            is_expected.to eq [award1, award2]
```
Alexey1100, line 33: Better use `.to match_array` to avoid flakiness.

spec/queries/next_blockchain_transactables_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+          let!(:prioritized_award) { FactoryBot.create :award, project: project, status: :accepted, prioritized_at: Time.zone.now }
+          let(:batch_size) { 100 }
+          let(:token) { FactoryBot.create :token, :ropsten }
+
+          it 'returns the first prioritized transfers only (ignore batch)' do
+            is_expected.to eq [prioritized_award]
+          end
+        end
+      end
+    end
```
Alexey1100, line 90: Looks like it needs more coverage for cancelled/failed transactions, as well as TransferRules and AccountTokenRecords, since logic there is a bit different.

spec/controllers/api/v1/blockchain_transactions_controller_spec.rb
Riddlerrr: Yeah, it must work, checking...

Riddlerrr: fixed

app/queries/next_blockchain_transactables.rb
Riddlerrr: I agree. Initially it was like this but it add cyclomatic complexity and rubocop was not happy =)
Ok, let's ignore rubocop

app/queries/next_blockchain_transactables.rb
Riddlerrr: 👍 Moved it into scopes_for_awards

app/queries/next_blockchain_transactables.rb
Alexey1100: Yeah, or we can just extract these two checks into a separate method like `should_be_processed?`

spec/factories/tokens.rb
```diff
@@ -25,5 +25,21 @@
...
+      batch_contract_address { '0x68ac9a329c688afbf1fc2e5d3e8cb6e88989e2cc' }
+    end
+
+    trait :lockup do
+      contract_address { '0x9608848FA0063063d2Bb401e8B5efFcb8152Ec65' }
+      batch_contract_address { nil }
+      _token_type { :token_release_schedule }
+      _blockchain { :ethereum_rinkeby }
+      decimal_places { 10 }
+      token_frozen { false }
```
Alexey1100, line 42: Looks like this line is not needed here

app/queries/next_blockchain_transactables.rb
```diff
@@ -0,0 +1,156 @@
...
+      (blockchain_transactions.id IS NULL) OR
+      (blockchain_transactions.status IN (:blockchain_transactions_cancelled_statuses)) OR
+      (blockchain_transactions.status = :blockchain_transaction_created_status AND blockchain_transactions.created_at < :timestamp)
+    SQL
+
+    q =
+      if transactable_class == Award
+        project.public_send(transactable_class.table_name)
+      else
+        project.token.public_send(transactable_class.table_name)
```
Alexey1100, line 52: Will not work properly with `Token` transactable, which is used for token freeze

spec/factories/accounts.rb
```diff
@@ -9,5 +9,13 @@
     trait :unconfirmed do
       email_confirm_token { SecureRandom.hex(6) }
     end
+
+    trait :verified do
+      verifications { [association(:verification)] }
+    end
+
+    trait :unverified do
+      verifications { [association(:verification, passed: false)] }
```
Alexey1100, line 18: I think it makes sense to call it `verification_failed`, since unverified is default state when verification doesn't exist.

spec/queries/next_blockchain_transactables_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+      end
+
+      context 'return transfer_rules with latest blockchain_transaction Failed' do
+        let!(:blockchain_transaction) { create(:blockchain_transaction_transfer_rule, status: :failed, blockchain_transactables: transfer_rule) }
+
+        it { is_expected.to include(transfer_rule) }
+      end
+    end
+  end
+end
```
Alexey1100, line 275: Looks like we need coverage for Token freeze as well here, since this logic is now used everywhere.

Also could you go through UI with both Metamask/WalletConnect and OREID, and verify that Transfers, AccountTokenRecords, TransferRules (creation/deletion) and Token freeze is not broken, since we don't have tests for that.

Alexey1100 (changes requested):

spec/queries/next_blockchain_transactables_spec.rb
Alexey1100: You can use the project currently present in seeds for testing:
` FactoryBot.create :project, :using_security_token, account: Account.first`

Let me know if you need permissions for the security token it's using.

app/controllers/api/v1/blockchain_transactions_controller.rb
```diff
@@ -8,7 +8,7 @@ class Api::V1::BlockchainTransactionsController < Api::V1::ApiController
   def create
     return head :no_content if creation_disabled?
 
-    if transactable
+    if transactable.any?
```
zhinfuhnya, line 11: As far as I understand `NextBlockchainTransactables` returns relation, in order to it better to use plural name.

app/queries/next_blockchain_transactables.rb
```diff
@@ -0,0 +1,156 @@
...
+    def hot_wallet_manual?
+      target_hot_wallet? && project.hot_wallet_manual_sending?
+    end
+
+    def hot_wallet_auto?
+      target_hot_wallet? && project.hot_wallet_auto_sending?
+    end
+
+    def hot_wallet_support?(transactable_class)
+      transactable_class.in?([Award, AccountTokenRecord])
```
zhinfuhnya, line 142: `transactable_class.in?(default_transactable_classes)`

zhinfuhnya (approved):
👨‍🚀 🙆‍♂️ 🚀 💫

app/models/concerns/blockchain_transactable.rb
```diff
@@ -10,71 +10,8 @@ module BlockchainTransactable
...
-    }
-
-    scope :ready_for_hw_manual_blockchain_transaction, lambda {
-      if table_name.in?(%w[awards account_token_records])
-        ready_for_blockchain_transaction(true).where("#{table_name}.prioritized_at is not null")
-      else
-        self.class.none
-      end
+    scope :ready_for_batch_blockchain_transaction, lambda { |project|
+      NextBlockchainTransactables.new(project: project, target: :manual, transactable_classes: [Award]).call
```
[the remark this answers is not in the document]
Riddlerrr: Let's just delete it  =)

app/queries/next_blockchain_transactables.rb
Riddlerrr: for now it is equal, yes but they can be different in a future and it is not related. So let's leave it as is

app/queries/next_blockchain_transactables.rb
```diff
@@ -0,0 +1,160 @@
...
+
+      transactions = next_transactables_for(transactable_class)
+      break if transactions.any?
+    end
+
+    transactions
+  end
+
+  def next_transactables_for(transactable_class, batch_size: 1) # rubocop:todo Metrics/CyclomaticComplexity
+    return Token.where(id: project.token_id) if transactable_class == Token
```
Alexey1100, line 30: This line and...

app/queries/next_blockchain_transactables.rb
```diff
@@ -0,0 +1,160 @@
...
+      (blockchain_transactions.id IS NULL) OR
+      (blockchain_transactions.status IN (:blockchain_transactions_cancelled_statuses)) OR
+      (blockchain_transactions.status = :blockchain_transaction_created_status AND blockchain_transactions.created_at < :timestamp)
+    SQL
+
+    q =
+      if transactable_class == Award
+        project.awards
+      elsif transactable_class == Token
+        project.token
```
Alexey1100, line 54: ...this one seem to be repeating each other

Alexey1100 (changes requested):

Alexey1100 (approved):

PR #2248, opened 2021-07-13
Transfer settings menu

zhinfuhnya:
https://www.pivotaltracker.com/story/show/178544187


Riddlerrr (approved):
👍

app/views/dashboard/transfers/_transfer.html.erb
```diff
@@ -103,24 +99,23 @@
...
-                        project_award_type_award_path(transfer.project, transfer.award_type, transfer),
-                        method: :delete,
-                        class: 'dropdown-item',
-                        role: 'button' %>
-          <% end %>
+        <div class="dropdown-menu">
+          <turbo-frame target="_top"
+                       loading="lazy"
+                       id="transfer_settings"
+                       src="<%= project_transfer_settings_path(transfer_id: transfer.id) %>">
```
Alexey1100, line 117: Oh looks like it's not in the ticket, but we discussed that before.

The important part is to make the menu lazy load each time, similar to transfer form, so the prioritise button will be always up-to-date.

I've added the details in the ticket.

Alexey1100 (changes requested):

Alexey1100 (approved):

PR #2249, opened 2021-07-14
Improve OREID payment integration

Alexey1100:
Resolves:
https://www.pivotaltracker.com/story/show/177615677
https://www.pivotaltracker.com/story/show/178529474
https://www.pivotaltracker.com/story/show/178529600
https://www.pivotaltracker.com/story/show/178883783

- Improve OREID stimulus controller
- Integrate OREID stimulus controller into Accounts and Transfer Rules
pages
- Add flash message regarding missing OREID wallet when redirecting to Wallets page
- Fix duplicated records on Accounts page

### DIFF
--- a/app/controllers/dashboard/accounts_controller.rb
+++ b/app/controllers/dashboard/accounts_controller.rb
@@ -129,7 +129,7 @@ class Dashboard::AccountsController < ApplicationController
       if @project.token&.token_type&.operates_with_account_records?
         @q = @q.joins(:account_token_records)
                .where(account_token_records: { token_id: @project.object.token_id })
-               .includes(account_token_records: [:reg_group])
+               .includes(account_token_records: [:reg_group]).distinct
       end
 
       @q = @q.ransack(params[:q])
--- a/app/controllers/dashboard/accounts_controller.rb
+++ b/app/controllers/dashboard/accounts_controller.rb
@@ -129,7 +129,8 @@ class Dashboard::AccountsController < ApplicationController
       if @project.token&.token_type&.operates_with_account_records?
         @q = @q.joins(:account_token_records)
                .where(account_token_records: { token_id: @project.object.token_id })
-               .includes(account_token_records: [:reg_group]).distinct
+               .includes(account_token_records: [:reg_group])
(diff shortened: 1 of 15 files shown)


app/controllers/dashboard/accounts_controller.rb
```diff
@@ -129,7 +129,7 @@ def accounts_query
       if @project.token&.token_type&.operates_with_account_records?
         @q = @q.joins(:account_token_records)
                .where(account_token_records: { token_id: @project.object.token_id })
-               .includes(account_token_records: [:reg_group])
+               .includes(account_token_records: [:reg_group]).distinct
```
zhinfuhnya, line 132: it's better to move `.distinct` to a new line : )

app/controllers/wallets_controller.rb
```diff
@@ -105,6 +106,10 @@ def wallets_collection
       end
     end
 
+    def set_flash
+      flash.now[:error] = "To sign transactions with ORE ID, please link your ORE ID wallet and make sure it's set as primary." if params[:ore_id_required]
```
zhinfuhnya, line 110: In my opinion, our controllers have a lot of callbacks, which is not good practice.

I'm in favor of bringing the flash directly into the action:

```
def index
    @wallets = wallets_collection.includes(balances: :token).order(:_blockchain)

    respond_to do |format|
      format.html { flash: { error: 'To sign transactions with ORE ID, please link your ORE ID wallet and make sure it's set as primary.' } if params[:ore_id_required] }
      format.json { render_collection_partial }
    end
  end
```

zhinfuhnya (dismissed):

app/controllers/wallets_controller.rb
Alexey1100: True, but it's all about balance :)
Params handlers should live in a callbacks I think.

aquabu: Let's discuss it and if we want to refactor we can include it in a later commit. The refactoring could be significant if we want to adopt a new pattern.

aquabu (approved):

PR #2251, opened 2021-07-15
Fix invitation logic issues

Alexey1100:
Resolves:
https://www.pivotaltracker.com/story/show/178890369
https://www.pivotaltracker.com/story/show/178890482
https://www.pivotaltracker.com/story/show/178890363
https://www.pivotaltracker.com/story/show/178890380

- Make reset passwords contoller available when invitation is required
- Allow email to match case insensitively when invite forces email
- Cleanup invite from session if accessing sign up page not from an invite
- Show interested projects when accessing from whitelabel


spec/models/account_spec.rb
```diff
@@ -31,6 +31,17 @@
       it { is_expected.to be_valid }
     end
 
+    context 'and email matches invite email' do
```
Alexey1100, line 34:
```suggestion
    context 'and email matches invite email in a different case' do
```

app/controllers/accounts_controller.rb
```diff
@@ -246,7 +247,11 @@ def account_decorate(account)
       )
     end
 
+    def cleanup_invite
+      session.delete(:invite_id) unless params[:invited]
```
zhinfuhnya, line 251: Perhaps it would be better to move the logic directly to action new instead of the callback? (so as not to overload the controller)

zhinfuhnya (approved):

app/controllers/accounts_controller.rb
Alexey1100: It should be executed before `:require_login_strict`

aquabu (approved):

PR #2264, opened 2021-07-21
Remove missions index page

zhinfuhnya:


config/routes.rb
```diff
@@ -1,8 +1,7 @@
 require 'sidekiq/web'
 
 Rails.application.routes.draw do
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
-  root 'pages#featured'
+  root 'projects#landing'
```
zhinfuhnya, line 4: What is the new root path?

zhinfuhnya:
@aquabu @Alexey1100 @Riddlerrr should we remove `featured` column from projects?

zhinfuhnya:
@aquabu @Alexey1100 we have two pages like `projects` and `projects/mine`. As far as I understand we have to redirect the user to the corresponding page depending on the presence of the user: an anonymous user to the projects, an authorized user to the landing?

Alexey1100 (approved):

PR #2266, opened 2021-07-22
Remove panoramic image from project form

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/178778462


zhinfuhnya:
@aquabu @Alexey1100 @Riddlerrr should we remove panoramic image column from projects table?

Alexey1100:
> @aquabu @Alexey1100 @Riddlerrr should we remove panoramic image column from projects table?

I think we should.
We might do it after the redesigning the project nav, so the image would not be used anywhere.

Alexey1100 (approved):

aquabu (approved):

PR #2267, opened 2021-07-22
Remove experiences, specialties and interests

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/178964759


zhinfuhnya:
@aquabu @Alexey1100 should we adapt the specialties to the roles of the project participants?

.overcommit.yml
```diff
@@ -51,17 +51,3 @@ PreCommit:
     enabled: true
     required_executable: 'bin/erblint_ci'
     command: ['bin/erblint_ci']
-
-# Hooks that run after HEAD changes or a file is explicitly checked out.
```
Riddlerrr, line 55: I believe it was committed accidentally

Riddlerrr (changes requested):
Please fix a failed specs

app/controllers/application_controller.rb
```diff
@@ -151,8 +151,10 @@ def task_to_props(task) # rubocop:todo Metrics/CyclomaticComplexity
     task&.serializable_hash&.merge({
       description_html: Comakery::Markdown.to_html(task.description),
       requirements_html: Comakery::Markdown.to_html(task.requirements),
-      specialty: task.specialty&.name,
-      mission: { name: task.project&.mission&.name },
+      mission: {
```
Alexey1100, line 154: Possible conflict with #2284

app/controllers/missions_controller.rb
```diff
@@ -152,6 +149,30 @@ def project_tokens(mission)
       }
     end
 
+    def set_mission_props # rubocop:todo Metrics/CyclomaticComplexity
```
Alexey1100, line 152: I don't think we need it after #2264

app/javascript/components/MyTask.jsx
```diff
@@ -489,7 +474,7 @@ class MyTask extends React.Component {
 
                   <Type>
                     <b>TYPE </b>
-                    {task.specialty || 'General'}
+                    {'General'}
                   </Type>
```
Alexey1100, line 478: I think we should remove the type, since it just indicates the specialty which is no longer present.

app/models/award.rb
```diff
@@ -251,6 +237,7 @@ def can_be_cloned_for?(account)
     (account.awards.started.count < STARTED_TASKS_PER_CONTRIBUTOR) && !reached_maximum_assignments_for?(account)
   end
 
+  # TODO: Should we remove assignments?
```
Alexey1100, line 240: Nope, it's different logic :)

public/doc/api/v1/index.html
```diff
@@ -282,18 +282,6 @@ <h2>VII. Blockchain Transactions</h2>
...
-        <a href="/doc/api/v1/vii._blockchain_transactions/generate_transaction.html">GENERATE TRANSACTION</a>
-      </li>
-      <li>
-        <a href="/doc/api/v1/vii._blockchain_transactions/generate_transaction_–_with_project_transaction_api_key.html">GENERATE TRANSACTION – WITH PROJECT TRANSACTION API KEY</a>
-      </li>
-      <li>
-        <a href="/doc/api/v1/vii._blockchain_transactions/generate_transaction_–_transactable_type.html">GENERATE TRANSACTION – TRANSACTABLE TYPE</a>
-      </li>
-      <li>
-        <a href="/doc/api/v1/vii._blockchain_transactions/generate_transaction_–_transactable_type_and_transactable_id.html">GENERATE TRANSACTION – TRANSACTABLE TYPE AND TRANSACTABLE ID</a>
```
Alexey1100, line 295: These links shouldn't be removed

public/doc/api/v1/vii._blockchain_transactions/generate_transaction.html
```diff
@@ -1,361 +0,0 @@
-<!DOCTYPE html>
-<html>
```
Alexey1100, line 2: These and any other docs should't be removed

Alexey1100 (changes requested):

Alexey1100 (approved):

Riddlerrr (commented):
It's pretty big changes, so maybe we should extract migrations to a separate PR and ensure we don't break anything before deleting data?

cc @Alexey1100 @aquabu

db/data_migrations/20210426121452_fill_roles_with_interests.rb
```diff
@@ -1,19 +0,0 @@
-class FillRolesWithInterests < ActiveRecord::DataMigration
```
Riddlerrr, line 1: This file was not reverted

spec/data_migrations/fill_roles_with_interests_spec.rb
```diff
@@ -1,33 +0,0 @@
-require 'rails_helper'
```
Riddlerrr, line 1: And this

db/data_migrations/20210426121452_fill_roles_with_interests.rb
Riddlerrr: Oh.. it will not work anyway because Interest was deleted

Riddlerrr (approved):

db/data_migrations/20210426121452_fill_roles_with_interests.rb
zhinfuhnya: No, you're right : )

It's also need to be reverted.

PR #2270, opened 2021-07-23
DatabaseCleaner: Use truncation strategy as a default

Riddlerrr:
task: https://www.pivotaltracker.com/story/show/178978314

I read docs and suggestions about strategies. It recommends to use `:transaction` everywhere it possible.
Also `transaction` strategy brings random deadlock errors, which can be a real problem with CI.

So, I setup `truncation` strategy for `tasks` specs and left `transaction` as a default strategy.

### DIFF
--- a/spec/support/database_cleaner.rb
+++ b/spec/support/database_cleaner.rb
@@ -3,7 +3,15 @@ RSpec.configure do |config|
     DatabaseCleaner.clean_with(:truncation)
   end
 
-  config.before(:each) do
+  config.before do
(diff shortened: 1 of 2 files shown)


Riddlerrr:
Interesting... It starts failing in different places.

Alexey1100:
Yeah, looks like lots of API-related specs are failing with invalid nonce.

Riddlerrr:
I updated PR's description

Alexey1100 (approved):

PR #2291, opened 2021-07-30
Require invitation whitelabel can not see sign up

sofiyareverse:
https://www.pivotaltracker.com/n/projects/2167210/stories/178970130


spec/features/pages/new_styling_spec.rb
```diff
@@ -22,8 +22,21 @@
...
 
+  context 'can not see sign up if whitelabel mission require invitation' do
+    let!(:whitelabel_mission) { create(:whitelabel_mission, whitelabel_domain: 'www.example.com', require_invitation: true) }
+
+    before do
+      visit styleguide_path
+    end
+
+    it 'returns vertical navbar without sign up when whitelabel mission requires invitation' do
+      expect(page).not_to have_css('span', text: 'Sign Up')
```
Riddlerrr, line 33: Let's check for "Sign in" exists before checking for Sign up is not there. To ensure page loaded correctly.

Riddlerrr (changes requested):

spec/features/pages/new_styling_spec.rb
sofiyareverse: for now we can not do this. It will exist if we have navbar on login page because for now navbar is only on styleguide page before the merge. after the merge it will be seen on session_new_path... and tests will be changed also for pages

Riddlerrr (approved):
👍

PR #2303, opened 2021-08-04
Configure overcommit

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/178965253


.overcommit.yml
```diff
@@ -37,6 +27,13 @@
 # other checks that you want to run before you allow a commit object to be
 # created.
 PreCommit:
+  RuboCop:
+    enabled: true
+    on_warn: fail
+  TrailingWhitespace:
+    enabled: true
+    exclude:
+      - '**/db/structure.sql'
```
Riddlerrr, line 36: I think we don't have structure.sql, we use `schema.rb` instead, so it can be deleted

Riddlerrr (approved):

Riddlerrr:
I've also moved rubocop config to default path to project root directory

Alexey1100 (changes requested):
Interesting, lots of files got autocorrected while config hasn't changed.
Is it working properly?

zhinfuhnya:
> Interesting, lots of files got autocorrected while config hasn't changed.
> Is it working properly?

@Alexey1100 in my case RuboCop works without autocorrection, it just points to the files that need to be corrected.

Alexey1100:
@zhinfuhnya we have changes checked in here which shouldn't be here, because the config hasn't changed.

zhinfuhnya:
@Alexey1100 Indeed, if I make a change for one line, RuboCop check on the whole file.

Riddlerrr:
I've updated rubocop_todo file so now it works good.
I've also enabled more checks.

Alexey1100:
Seems like what I'm seeing is related to:

https://github.com/sds/overcommit/issues/704

Alexey1100:
Fix above is not yet released, so resolved by manually adding `--disable-pending-cops` flag to overcommit rubocop config.

Alexey1100 (approved):

PR #2297, opened 2021-08-03
Password Change

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/179092377

PR #2305, opened 2021-08-04
Password Change

zhinfuhnya:
https://www.pivotaltracker.com/n/projects/2167210/stories/179092377

Rework https://github.com/CoMakery/comakery-server/pull/2297


app/controllers/password_resets_controller.rb
```diff
@@ -44,10 +45,16 @@ def update
 
   private
 
-    def permitted_param
+    def account_params
       params.require(:account).permit(:password)
     end
 
+    def mitigate_session_fixation
```
Riddlerrr, line 52: Let's extract it to a concern because we already use it in second time

Riddlerrr (approved):

Riddlerrr (approved):

PR #2311, opened 2021-08-05
UPS-81 Project modal form

sofiyareverse:
https://comakery-upside.atlassian.net/jira/software/projects/UPS/boards/4?selectedIssue=UPS-81


app/controllers/projects_controller.rb
```diff
@@ -75,6 +75,34 @@ def create
     end
   end
 
+  def modal_create
```
zhinfuhnya, line 78: Please stick RESTFul best practices, one of them — controllers should only respond to actions:            
    `%i[ index create update destroy ]`

For example, if this modal is responsible for quick setup of the project arrange the routes like:
```
    resources :projects do
      resource :quick_setups, only: [:create], controller: 'projects/quick_setups'
    end
```

Also, put all the logic inside the action in `interactors`. Like `Projects::QuickSetup`. Don't forget to add specifications for the interactor.

app/javascript/src/dist/css/project_modal.css
```diff
@@ -0,0 +1,28 @@
+
```
zhinfuhnya, line 1: Move css to `app/javascript/src/dist/css/projects/quick_setups/modal.css`

app/views/projects/_modal.html.erb
```diff
@@ -0,0 +1,93 @@
...
+            <hr>
+            <div class="pb-3">
+              <div class="d-inline project-modal-subtitle pr-16 h3">Invite team members to help with project setup</div>
+              <div class="d-inline project-modal-sub py-1 px-2 rounded project-modal-sub-grey-color">OPTIONAL</div>
+            </div>
+            <div class="hidden">
+              <div class="row row-cards pb-3" data-project-target='inviteAccountForm'>
+                <div class="col-lg-6">
+                  <label class="form-label">First Name</label>
+                  <%= text_field_tag "#{f.object.class.to_s.underscore}[first_names][]", nil, class: 'form-control', placeholder: '' %>
```
zhinfuhnya, line 57: Why do we need this construction?

`f.object.class.to_s.underscore`

spec/features/projects/new_modal_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    find('[data-target="#create-project"]').click
+    expect(page).to have_content('Quick Project Setup')
+
+    fill_in 'project[title]', with: Faker::Lorem.word
+    select 'Member', from: 'project[visibility]'
+    select 'No Token', from: 'project[token_id]'
+
+    click_on 'Publish'
+
+    sleep 2
```
zhinfuhnya, line 23: Without sleep doesn't work?

Capybara `have_` methods includes page waiting time.

zhinfuhnya (changes requested):

spec/features/projects/new_modal_spec.rb
sofiyareverse: doesn't work... can't find page object

app/views/projects/_modal.html.erb
sofiyareverse: yep, you are right, we don't need to user project param for that (param[:project][:emails]), just param[:emails]

app/controllers/projects/quick_setups_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  def create
+    @project = current_account.projects.build project_params
+    @project.long_id ||= SecureRandom.hex(20)
+    @project.mission = @whitelabel_mission if @whitelabel_mission
+
+    authorize @project
+
+    respond_to do |format|
+      if @project.save
+        params[:emails].each_with_index do |account_email, i|
```
Riddlerrr, line 15: Let's extract what that after save logic to a separate private method. `add_role_or_send_invite` for example

app/decorators/project_decorator.rb
```diff
@@ -303,4 +303,8 @@ def ratio_pretty(value, total)
       "≈ #{ratio} %"
     end
   end
+
+  def self.visibility_options_for_select
+    Project.visibilities.keys.map { |name| [name.humanize, name] }
```
Riddlerrr, line 308: I think `member` should renamed to 'Private'
@aquabu Am I right?

app/javascript/controllers/project_controller.js
```diff
@@ -0,0 +1,13 @@
+import { Controller } from 'stimulus'
+
+export default class extends Controller {
+  static targets = [ 'inviteAccountForm', 'addAccountForm' ]
+
+  connect() {
+    this.addAccountFormTarget.appendChild(this.inviteAccountFormTarget.cloneNode(true))
```
Riddlerrr, line 7: Same body for both functions. Let's extract it to a separate function and reuse it in both places.

app/controllers/projects/quick_setups_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+          if account
+            ProjectRole.create(account: account, project: @project, role: account_role)
+          else
+            SendInvite.call(project: @project,
+                            whitelabel_mission: @whitelabel_mission,
+                            params: { email: account_email, role: account_role })
+          end
+        end
+
+        format.json { render json: { project_id: @project.id, whitelabel: @project.mission }, status: :created }
```
Riddlerrr, line 28: We should render `turbo-frame` here, not json.
You can see example here: https://github.com/CoMakery/comakery-server/blob/075688fed9a0c9d7d5335f56b83aa94c1224130f/app/views/dashboard/transfers/edit.html.erb#L1-L3

app/javascript/controllers/projects/form_controller.js
```diff
@@ -0,0 +1,58 @@
...
+
+  connect() {
+    this.createdEvent = new CustomEvent('invites:created', { bubbles: true });
+    this.jmodal = this.element.jModalController;
+  }
+
+  onSubmit(event) {
+    event.preventDefault();
+
+    Rails.ajax({
```
Riddlerrr, line 17: After adding a response with `turbo-frame` you not need all that complex js and error handling

spec/features/projects/quick_setup_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    find('[data-target="#create-project"]').click
+    expect(page).to have_content('Quick Project Setup')
+
+    fill_in 'project[title]', with: Faker::Lorem.word
+    select 'Member', from: 'project[visibility]'
+    select 'No Token', from: 'project[token_id]'
+
+    click_on 'Publish'
+
+    sleep 2
```
Riddlerrr, line 23: It should work if you first check for `have_content` and only after that `have_current_path`.

`have_content` will wait for page load.

sleep should be removed of course.

Riddlerrr (changes requested):
The main change required is `turbo-frame` response instead of json.

app/controllers/projects/quick_setups_controller.rb
sofiyareverse: I think I'll add interactor like Bogdan proposed, it seems like a good idea

spec/features/projects/new_modal_spec.rb
Alexey1100: Yes, we shouldn't use sleep.
Pls use finders and expects to wait for the element to appear on the page.

PR #2314, opened 2021-08-06
UPS-10 Admins can preview token vestings for a TokenLockup transfer from the transfers page

Alexey1100:
- Create VestingSchedule model
- Create VestingSchedule controller and define routes


app/controllers/vesting_schedules_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  def show
+    vesting_schedule
+    amount
+    commencement_date
+  end
+
+  private
+
+    def token
+      @token ||= Token.find(params[:token_id])
```
zhinfuhnya, line 14: There seems to be no need to use memoization in all actions because we do not need to reuse instances per one request.

I also do not see the need to put instance variables in separate methods (that's optional).

zhinfuhnya (changes requested):
Good job :)

PR #2310, opened 2021-08-05
[UPS-94] Transfer rules page redesign

Riddlerrr:

PR #2321, opened 2021-08-13
Transfer rules: Create transfer rule modal

Riddlerrr:
https://comakery-upside.atlassian.net/browse/UPS-86


Riddlerrr:
Merged changes into #2310

